PR #45, opened 2024-09-06
Migrate query api to rust

ruokun-niu:
# Description

Migrate Query API to Rust

## Type of change

Fixes: #issue_number


infrastructure/comms-abstractions/src/comms.rs
```diff
@@ -36,5 +36,5 @@ pub trait Publisher {
         &self,
         data: Value,
         headers: Headers,
-    ) -> Result<(), Box<dyn std::error::Error>>;
+    ) -> Result<reqwest::Response, Box<dyn std::error::Error>>;
```
danielgerlag, line 39: I don't think we should include `reqwest::Response` as part of the API of drasi comms.  It tightly couples us to a specific library.

sources/shared/query-api/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    let mut headers_map = std::collections::HashMap::new();
+    _ = match headers.get("traceparent") {
+        Some(tp) => match tp.to_str() {
+            Ok(tp) => {
+                headers_map.insert("traceparent".to_string(), tp.to_string());
+            }
+            Err(e) => {} // 
+        },
+        None => {}
```
danielgerlag, line 113: We could log a warning if the trace headers are missing.

danielgerlag (changes requested):

infrastructure/comms-abstractions/src/comms.rs
ruokun-niu: How about a custom struct that contains the status code and/or the response body? Something like:
```rust
pub struct HttpResponse {
    pub status_code: u16,
    pub body: String,
}
```

danielgerlag: That might make sense for the `Invoker`, but the publisher implementation could be one that does not use Http, if we choose to not use Dapr in the future, many messages brokers have their own protocols.
I think if a non success code comes from the Dapr sidecar, the publish was unsuccessful, and we just return an error, and if the body contains details, we include that in the error result.

ruokun-niu: that makes sense. for a successful service invocation, maybe we can just return the body as Bytes?

ruokun-niu: oh actually this is the publisher, we don't really need to return anything if the publish is successful

ruokun-niu: Sorry I was think about the wrong thing

infrastructure/comms-abstractions/src/comms.rs
```diff
@@ -26,7 +26,7 @@ pub trait Invoker {
         app_id: String,
         method: String,
         headers: Headers,
-    ) -> Result<(), Box<dyn std::error::Error>>;
+    ) -> Result<serde_json::Value, Box<dyn std::error::Error>>;
```
danielgerlag, line 29: We can't assume the response will always be json

danielgerlag (approved):

PR #90, opened 2024-10-11
Resource Cleanup on delete

danielgerlag:
# Description

This PR adds infrastructure to clean up resources when sources / reactions are deleted.  It takes a 2 pronged approach.

The first is to add an optional `deprovisionHandler` to any resource service.  This is defined on the Source / Reaction Provider spec. When the resource is deleted, the control plane will do a best effort attempt to call the deprovision method on the service, using standard Dapr retry semantics. If the handler is not callable, to resource will still be deleted.

To second, is to include an `INSTANCE_ID` environment variable on each service.  This value remains the same for the lifetime of the resource (source/reaction). If you delete and recreate the resource with the same name, the instance id will change. Thus the service can inspect if the instance id is as expected, if not then any existing state should be considered stale.

Both mechanisms have been implemented for the debezium reactivator.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -121,7 +121,7 @@ jobs:
       - name: Build Management API
         uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5.4.0
         with:
-          context: ./control-planes
+          context: '.'
           file: ./control-planes/mgmt_api/Dockerfile
           platforms: linux/amd64, linux/arm64
           tags: |
--- a/cli/service/resources/default-source-providers.yaml
+++ b/cli/service/resources/default-source-providers.yaml
@@ -15,6 +15,9 @@ spec:
             default: pg
     reactivator: 
       image: source-debezium-reactivator
+      deprovisionHandler: true
+      dapr:
+        app-port: "80"
       config_schema:
         type: object
         properties:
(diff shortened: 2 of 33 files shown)


ruokun-niu (approved):

ruokun-niu:
Should we set `deprovisionHandler` to be true for all sources?

ruokun-niu:
Quick question (I might be jumping ahead since I vaguely remember you mentioning that there will be multiple PRs for source cleanups): If I set `deprovisionHandler` to true for one of my source reactivator, will the dapr resources for other components of the source be deleted when the source is deleted? For example, when I delete the source, will the dapr state data of the change router be cleared?

danielgerlag:
> Should we set `deprovisionHandler` to be true for all sources?

Only once we have implemented handlers for them

danielgerlag:
> Quick question (I might be jumping ahead since I vaguely remember you mentioning that there will be multiple PRs for source cleanups): If I set `deprovisionHandler` to true for one of my source reactivator, will the dapr resources for other components of the source be deleted when the source is deleted? For example, when I delete the source, will the dapr state data of the change router be cleared?

It's per service. Since change router is a universal service across all sources, if we implemented a handler, it wouldn't need a special config, so we could to call it if we wanted to.

agentofreality:
I think, in future, we should do separate PRs for things like the changes to the shared infrastructure abstractions.

agentofreality (approved):

PR #94, opened 2024-10-18
Updated change router to retain subscription after a pod restart

ruokun-niu:
# Description

- The change router was unable to retrieve the subscription info from the state store because the subscription values were not properly saved in the state store. This PR created a custom struct called `StateManager` that is responsible for saving state values to the dapr state store. The `save_state` function from the Dapr Rust SDK will save the state as a serialized string, which prohibits the `query_state_alpha1` function to fetch the state result.
  - For the save_state function in the StateManager, the state value should be an array of json object, where each object needs to have a `key` field and a `value` field

## Type of change

Fixes:[ #issue_number](https://dev.azure.com/azure-octo/Incubations/_boards/board/t/Drasi/User%20Stories?workitem=13392)


sources/shared/change-router/src/state_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+                        Value::Object(obj) => {
+                            if !obj.contains_key("key") || !obj.contains_key("value") {
+                                return Err(Box::from("State entry object must have 'key' and 'value' fields"));
+                            }
+                        }
+                        _ => return Err(Box::from("State entry must be a JSON array of objects")),
+                    }
+                }
+            }
+            None => return Err(Box::from("State entry must be a JSON array")),
```
danielgerlag, line 49: Why don't we just make the `state_entry` parameter a `Vec<Value>`?

sources/shared/change-router/src/state_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        // check if the state entry is a json array
+        match state_entry.as_array() {
+            Some(entry) => {
+                // check each entry, if it is a string continue
+                // if it is an object, check if it has a "key" field and a "value" field
+                for entry in entry {
+                    match entry {
+                        Value::Object(obj) => {
+                            if !obj.contains_key("key") || !obj.contains_key("value") {
```
danielgerlag, line 41: I think a strongly type object would be better here

sources/shared/change-router/src/state_manager.rs
ruokun-niu: Good point. Also I checked Dapr's documentation and their version of `save_state` accepts the `state_entry` as `Vec<u8>`. Should we be consistent with them? 
I guess the downside of using Vec<u8> is that we still need to deserialize the data into `Value` to check if the required fields are missing. Maybe it is better to use Vec<Value> for this

sources/shared/change-router/src/state_manager.rs
```diff
@@ -0,0 +1,77 @@
...
+            client: reqwest::Client::new(),
+            dapr_host: dapr_host.to_string(),
+            dapr_port,
+            store_name: store_name.to_string(),
+        }
+    }
+
+    pub async fn save_state(
+        &self,
+        state_entry: Vec<Value>,
```
danielgerlag, line 32: Could the API for the state manager take the strongly typed object, so that there is no need to validate for specific fields in an untyped object?

sources/shared/change-router/src/main.rs
```diff
@@ -37,20 +39,25 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            "EQ": { "type": "SourceSubscription"}
-        },
+            "EQ": { "type": "SourceSubscription" }
+        }
     });
+    let mut metadata = std::collections::HashMap::new();
+    metadata.insert("contentType".to_string(), "application/json".to_string());
     let response = match dapr_client
-        .query_state_alpha1(config.clone().subscriber_store, query_condition, None)
+        .query_state_alpha1(
```
danielgerlag, line 48: should we encapsulate both the reading and writing of state in the state_manager?

ruokun-niu: Yeah that makes sense. Also we will have a deprovision function in the future for deleting the subscription from the state so it should also go into the same place

agentofreality (approved):

PR #103, opened 2024-10-30
Review StoredProc Reaction

ruokun-niu:
# Description

- Updated storedproc reaction provider
- StoredProc now accepts three different comands: AddedResultCommand, UpdatedResultCommand and DeletedResultCommand
- Added comments
- Created a test for the storedproc reaction in the `e2e` folder
  - The storedproc test scenario uses the same postgres database as the simple-scenario, but it uses its own source, queries and reaction

## Type of change

Fixes: (https://dev.azure.com/azure-octo/Incubations/_boards/board/t/Drasi/User%20Stories?workitem=12558)


reactions/sql/storedproc-reaction/src/index.ts
```diff
@@ -0,0 +1,184 @@
...
+      user :   databaseUser,
+      password : databasePassword,
+      database : databaseDbname,
+      ssl: databaseSsl
+    }
+  });
+
+
+const queryParamsRegex = /@\w+/g;
+const addedResultCommand: string = process.env["AddedResultCommand"] ?? '';
```
danielgerlag, line 41: You could use `getConfigValue` instead of `process.env`

reactions/sql/storedproc-reaction/src/index.ts
```diff
@@ -0,0 +1,184 @@
...
+        for (const match of matches) {
+            const param: string = match.substring(1);
+            deletedResultCommandParamList.push(param);
+        }
+    }
+}
+
+let storedProcReaction = new DrasiReaction(onChangeEvent,{
+    parseQueryConfig: parseYaml,
+    onControlEvent: onControlEvent
```
danielgerlag, line 85: We are not using per query configs, so we could remove `parseQueryConfig: parseYaml`
We could also remove the control event handler if we are not doing anything in it.

reactions/sql/storedproc-reaction/src/index.ts
```diff
@@ -0,0 +1,184 @@
...
+      } else {
+        query += queryArguments[i];
+      }
+      query += (i < queryArguments.length - 1) ? ", " : "";
+    }
+    query += ")";
+    console.log(`Executing the stored proc: ${query}`);
+  
+    // Execute the query
+    knex.raw(query).then(() => {
```
danielgerlag, line 168: you could make this function `async` and use `await`

e2e-tests/fixtures/storedproc-fixture.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @param {Array} queryIds
+ */
+function storedprocReactionManifest(queryIds) {
```
danielgerlag, line 4: this could be pure yaml in the directory of your specific scenario

e2e-tests/simple-scenario/my.test.js
```diff
@@ -62,4 +63,24 @@ test('Initial data', async () => {
   let initData = await signalrFixture.requestReload("query1");
 
   expect(initData.length == 2).toBeTruthy();
-});
\ No newline at end of file
+});
+
+
+test('Test StoredProc Reaction - AddedResultCommand', async () => {
```
danielgerlag, line 69: Let's create a new scenario instead of extending the simple scenario.

danielgerlag (changes requested):

e2e-tests/simple-scenario/resources.yaml
```diff
@@ -29,6 +29,28 @@ data:
 
     INSERT INTO "Item" ("ItemId", "Name", "Category") VALUES (1, 'Foo', '1');
     INSERT INTO "Item" ("ItemId", "Name", "Category") VALUES (2, 'Foo', '1');
+
+
+
+    CREATE TABLE "CommandResult" (
```
danielgerlag, line 35: Do we still need this?

e2e-tests/storedproc-scenario/storedproc.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+test('Test StoredProc Reaction - AddedResultCommand', async () => {
+  // add a new item
+  const storedprocResources = yaml.loadAll(fs.readFileSync(__dirname + '/storedproc-reaction.yaml', 'utf8'));
+  await deployResources(storedprocResources);
+
+  await dbClient.query(`INSERT INTO "Item" ("ItemId", "Name", "Category") VALUES (3, 'Drasi', '1')`);
+
+  // sleep 35 seconds
+  await new Promise(r => setTimeout(r, 15000));
```
danielgerlag, line 51: Is there a better way we can do this rather than sleeping 35 seconds?
Could we check every 1 second and timeout after 30 seconds?
Something like the `.waitForChange` on on the signalr fixture.

reactions/sql/storedproc-reaction/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+// Helper functions
+
+// Define the function that will be called when a change event is received
+async function onChangeEvent(event: ChangeEvent): Promise<void> {
+    for (let added of event.addedResults) {
+        console.log(`Processing added results`);
+        const queryArguments: any[] = [];
+        if (checkSqlCommandParameters(added, addedResultCommandParamList, queryArguments)) { // checks if the results contain the parameters
+            console.log(`Issuing added command: ${addedResultCommand}`);
+            executeStoredProcedure(addedResultCommand, queryArguments);
```
danielgerlag, line 99: I think you need and `await` here

reactions/sql/storedproc-reaction/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+    query += ")";
+    console.log(`Executing the stored proc: ${query}`);
+  
+    // Execute the query
+    try {
+        await knex.raw(query);
+        console.log("The query was executed successfully");
+      } catch (error) {
+        console.log(error);
```
danielgerlag, line 168: If it fails, we probably want the exception to bubble up to the event handler, so that it will return a 500 error back to DAPR, so that the message will not be acknowledged off the queue.

danielgerlag (commented):
You might also need to look into doing some kind of teardown per scenario, before it starts the next scenario.

e2e-tests/simple-scenario/my.test.js
ruokun-niu: I investigated this a bit more and found out that jest actually does not allow two test scenarios to run sequentially. This sometimes caused a race condition in the postgres database and the e2e test to fail. 

I think maybe it is better to move the setup of the postgres database outside of the individual test scenarios and include it as a part of the cluster setup process?

e2e-tests/simple-scenario/resources.yaml
ruokun-niu: The storedprocedure adds an item to this table. This is used to verify that the storedproc reaction is working

danielgerlag: but this is on the wrong scenario?

e2e-tests/simple-scenario/my.test.js
danielgerlag: It would be better if scenarios did not share things. We could use different DNS names for different PG instances. I think you can run tests serially with jest: https://jestjs.io/docs/cli#--runinband

ruokun-niu: Working on it right now

cli/service/resources/default-reaction-providers.yaml
```diff
@@ -164,20 +164,26 @@ spec:
   config_schema: 
     type: object
     properties: 
-      SqlCommand:
+      AddedResultCommand:
```
danielgerlag, line 167: In all of our sources, we have adopted camelCase for properties. Perhaps we should standardize?

e2e-tests/02-storedproc-scenario/storedproc.test.js
```diff
@@ -0,0 +1,92 @@
...
+  await deployResources(resources);
+
+  dbClient.port = await dbPortForward.start();
+  await dbClient.connect();
+  await new Promise(r => setTimeout(r, 15000)); 
+}, 120000);
+
+
+afterAll(async () => {
+  await dbClient.end();
```
danielgerlag, line 45: I think we need to clean up resources after each scenario.
This is missing from the test infrastructure.

danielgerlag (commented):
Looking good

danielgerlag (approved):

agentofreality (approved):

PR #105, opened 2024-11-01
Updated the install message for a local install

ruokun-niu:
# Description

If `drasi init` is run with the `local` flag, the following message will be logged:

```bash
Installing Drasi version <version> with local images
```

## Type of change

Fixes: [#issue_number](https://dev.azure.com/azure-octo/Incubations/_boards/board/t/Drasi/User%20Stories?workitem=13255)


cli/cmd/init.go
agentofreality, line 1: @ruokun-niu does it still use a specific version if local install is active?

ruokun-niu: if the `local` flag is set, it will use the locally available images which does not have any versioning.

However, I just realized there is an edge case that is not currently covered: if the user executes the init command with both the version and local flags (e.g. drasi init --version 0.1.4 --local).

Maybe we should just throw an error if both flags are set? or we can make it so that the `local` flag to take precedence over the `version` flag
@agentofreality  @danielgerlag

danielgerlag: local images could be built with different version tags, but not by default

ruokun-niu: Ohh right. I'll update this message.

agentofreality (approved):

danielgerlag (approved):

PR #111, opened 2024-11-07
Review result reaction

ruokun-niu:
# Description

The result reaction is used to retrieve the current result set, or the result set at a given timestamp.

To retrieve the current result set, send a GET request to the endpoint `/{query_id}`, where `query_id` is the query you are interested in.

For example:
```bash
curl -X GET "localhost:8080/hello-world-from"
```

To retrieve the result set at a given timestamp, send a GET request to the endpoint `/{query_id}/{timestamp}`, where `query_id` is the query you are interested in, and `timestamp` is the timestamp you are interested in. (This requires the retentionPolicy of the continuous query to be set to `all` or `expire`
For example:
```bash
curl -X GET "localhost:8080/hello-world-from/123456789"
```
Sample output can be found here: https://gist.github.com/ruokun-niu/f5be829bdc6c82e332e99725c923a5e2

## Type of change

Fixes:[ #issue_number](https://dev.azure.com/azure-octo/Incubations/_boards/board/t/Drasi/User%20Stories?workitem=12555)


reactions/platform/result-reaction/Program.cs
```diff
@@ -26,50 +25,24 @@
...
-            result.Add(data);
-        }
-    }
-    Console.WriteLine("Result:" + result.Last());
-    return result.Last();
-});
 
 // Adding an endpoint that supports retrieving all results
-app.MapGet("/{queryId}/all", async (string queryId) => 
+app.MapGet("/{queryId}", async (string queryId) =>
```
danielgerlag, line 42: You could make this support large results really easily by adding support for streaming results, but return an `IAsyncEnumerable` - https://anthonygiretti.com/2021/09/22/asp-net-core-6-streaming-json-responses-with-iasyncenumerable-example-with-angular/

ruokun-niu: Hi @danielgerlag , I have updated the endpoints to return an `IAsyncEnumerable`, but I realized that if we just return this it will also return the metadata. I think that sometimes users would just want to get the data (for example in the quick test), so I have created two additional routes: `/{queryId}/data` and `/{queryId}/{ts}/data` that just returns the data as `IAsyncEnumerable<JsonElement>`. 
```bash
ruokunniu ~/Documents/playground  $ curl -X GET "localhost:8080/hello-world-from/12323423423232/data"
[{"MessageFrom":"Brian Kernighan","MessageId":2},{"MessageFrom":"Allen","MessageId":5},{"MessageFrom":"Allen","MessageId":6},{"MessageFrom":"Allen","MessageId":7}]%                                                                    
ruokunniu ~/Documents/playground  $ curl -X GET "localhost:8080/hello-world-from/12323423423232"     
[{"header":{"sequence":10,"timestamp":1731703800526,"state":"running"}},{"data":{"MessageFrom":"Brian Kernighan","MessageId":2}},{"data":{"MessageFrom":"Allen","MessageId":5}},{"data":{"MessageFrom":"Allen","MessageId":6}},{"data":{"MessageFrom":"Allen","MessageId":7}}]%                                                                             
ruokunniu ~/Documents/playground  $ curl -X GET "localhost:8080/hello-world-from"               
[{"header":{"sequence":10,"timestamp":1731703800526,"state":"running"}},{"data":{"MessageFrom":"Brian Kernighan","MessageId":2}},{"data":{"MessageFrom":"Allen","MessageId":5}},{"data":{"MessageFrom":"Allen","MessageId":6}},{"data":{"MessageFrom":"Allen","MessageId":7}}]%                                                                             
ruokunniu ~/Documents/playground  $ curl -X GET "localhost:8080/hello-world-from/data"
[{"MessageFrom":"Brian Kernighan","MessageId":2},{"MessageFrom":"Allen","MessageId":5},{"MessageFrom":"Allen","MessageId":6},{"MessageFrom":"Allen","MessageId":7}]%      
```

ruokun-niu: Thoughts?

ruokun-niu: I'll update the quick test based on this

reactions/platform/result-reaction/result-reaction.csproj
```diff
@@ -12,8 +12,8 @@
   <ItemGroup>
     <PackageReference Include="Dapr.Actors" Version="1.10.0" />
     <PackageReference Include="Dapr.Actors.AspNetCore" Version="1.9.0" />
-    <PackageReference Include="Dapr.AspNetCore" Version="1.10.0" />
-    <PackageReference Include="Dapr.Client" Version="1.10.0" />
+    <PackageReference Include="Dapr.AspNetCore" Version="1.14.0" />
+    <PackageReference Include="Dapr.Client" Version="1.14.0" />
```
danielgerlag, line 16: I think we can remove all the Dapr libraries.

reactions/platform/result-reaction/Program.cs
```diff
@@ -12,104 +12,117 @@
...
+using System.Net.Http;
+using ResultReaction.Services;
+using System.Text.Json;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+var builder = WebApplication.CreateBuilder(args);
+var configuration = BuildConfiguration();
+
+var queryContainerId = configuration["QueryContainerId"] ?? "default";
```
danielgerlag, line 28: Is this one of the reaction config properties?

ruokun-niu: Since the result reaction does not use the NuGet Reaction SDK, I did not modify how it retrieves the query container ID.

danielgerlag: I think this was originally a hack. If we keep it, we need to document this as a config property and that only queries within that container would work.

reactions/platform/result-reaction/Program.cs
```diff
@@ -12,104 +12,114 @@
...
+			yield return item;
+		}
+
+	}
+
+	return GetCurrentResult();
+});
+
+// Adding an endpoint that supports retrieving all results, returning only the data field
+app.MapGet("/{queryId}/data", async (string queryId) =>
```
danielgerlag, line 58: What happens if they request the results for a query not listed in the reaction config?

ruokun-niu: Good question. Since we are directly talking to the view service, I think they will still be able to retrieve the result of that query, even though it is not listed in the config. Should we do a check to ensure that the queryId is listed in the Reaction config

reactions/platform/result-reaction/Program.cs
```diff
@@ -12,104 +12,114 @@
...
+var queryContainerId = configuration["QueryContainerId"] ?? "default";
+
+
+builder.Services.AddSingleton<IResultViewClient, ResultViewClient>();
+
+
+var app = builder.Build();
+app.UseRouting();
+
+app.Urls.Add("http://0.0.0.0:80");  //dapr
```
danielgerlag, line 34: We should have the dapr endpoint and query endpoint hosted by 2 different web servers within the app.  The query endpoint will be exposed to the public, and if you call it with the right routes, you can pretend to be the dapr side car.

reactions/platform/result-reaction/Program.cs
danielgerlag: yes

reactions/platform/result-reaction/Program.cs
```diff
@@ -31,15 +39,23 @@
 var app = builder.Build();
 app.UseRouting();
 
-app.Urls.Add("http://0.0.0.0:80");  //dapr
 app.Urls.Add("http://0.0.0.0:8080"); //app
 
+// Dapr server
+var daprBuilder = WebApplication.CreateBuilder(args);
+var daprApp = daprBuilder.Build();
```
danielgerlag, line 46: Which app side dapr APIs are we implementing for this reaction?

ruokun-niu: Hmm I don't think we are implementing anything for this reaction? Was there something specific that stood out to you?

danielgerlag: If we aren't implementing any app side dapr APIs, then we don't need to have an open port for it at all?

ruokun-niu: Right good point. I just removed this and pointed dapr to the app port 8080

danielgerlag: i think we can just remove the dapr app port?

ruokun-niu: Yep that's what I did. I meant that I had to configure the port in the reaction provider, otherwise the sidecar container will fail to start

danielgerlag: I think if no app port is specified, then the sidecar will not try to connect to the app, and the sidecar should always start?

ruokun-niu: I think the issue with not setting an app port is that by default port 80 is used as the app port in the k8s provider (https://github.com/drasi-project/drasi-platform/blob/main/control-planes/kubernetes_provider/src/spec_builder/reaction.rs#L155).

One possible approach is to update this section of code and use `None` as the default value (indicating that we do not want sidecar to be enabled for this particular reaction). Although this might require changes in other reactions/reactionproviders. Perhaps this could be a separate task @danielgerlag ?

danielgerlag: Good point, I think that's out of scope for this PR... let's circle back later.

danielgerlag (approved):

PR #118, opened 2024-11-22
Initial Python Reaction SDK

sjrusso8:
# Reaction SDK for Drasi

This pull request provides the initial setup for implementing a [Drasi](https://drasi.io/) Reaction in Python.

## Background

This sdk uses generated [Pydantic models](https://docs.pydantic.dev/latest/concepts/models/) created from the spec laid out under `typespec/`. 

There is a bug with the tool used to create the generated model files where an import path might be like `import . from ChangeData`. [#2039](https://github.com/koxudaxi/datamodel-code-generator/issues/2039). These were fixed manually for now.

The sdk uses the [FastAPI Dapr extension](https://docs.dapr.io/developing-applications/sdks/python/python-sdk-extensions/python-fastapi/) to subscribe to events.  The overall code is similar to the already created sdks for javascript and dotnet. 

## Considerations

1. Not sure how the drasi team wants to publish the python package

### Install the package

```sh
pip install drasi_reaction_sdk
```

### Basic Example

```python
from drasi.reaction.models.ChangeEvent import ChangeEvent
from drasi.reaction.sdk import DrasiReaction


async def change_event(data: ChangeEvent, query_configs: dict[Any, Any] | None = None):
    print("handling change event")
    print(data)


reaction = DrasiReaction(on_change_event=change_event)

reaction.start()
```

### Advanced example

The following example illustrates 
 - Retrieving a configuration value from the Reaction manifest
 - Parsing the per query configuration object from Yaml
 - Process change events from the query
 - Process control events (start, stop, etc.) from the query

```python
from drasi.reaction.models.ChangeEvent import ChangeEvent
from drasi.reaction.models.ControlEvent import ControlEvent
from drasi.reaction.sdk import DrasiReaction 
from drasi.reaction.utils import parse_yaml, get_config_value

# Retrieve the connection string from the Reaction configuration
my_connection_string = get_config_value("MyConnectionString")

# Define the function that will be called when a change event is received
async def on_change_event(event: ChangeEvent, query_config: dict[str, Any] | None) -> None:
    print(f"Received change signal for query {event.query_id}")
    print(f"Passed in the query config: {query_config}")
    
    # do something else with the `event` and `query_config`

# Define the function that will be called when a control event is received
async def on_control_event(event: ControlEvent, query_config: dict[str, Any] | None) -> None:
    print(f"Received control signal: {event.control_signal} for query {event.query_id}")
    print(f"Passed in the query config: {query_config}")

print(f"Starting Drasi reaction with connection string: {my_connection_string}")

# Configure the Reaction with the on_change_event and on_control_event functions
custom_reaction = DrasiReaction(
    on_change_event=on_change_event,
    on_control_event=on_control_event,
    parse_query_configs=parse_yaml,
)

# Start the Reaction
custom_reaction.start()
```

## Testing

There are a few small unit tests to validate that a query is read from a directory, and subscribed to with dapr. 

Additional manual tests were ran with a local drasi instance using the "simple" & "advanced" examples in this change.


reactions/sdk/python/drasi_reaction/sdk.py
```diff
@@ -0,0 +1,145 @@
...
+    def start(self):
+        """Starts the application by subscribing to queries and running the FastAPI application."""
+
+        try:
+            self.subscribe()
+            logger.info("starting python reaction app")
+            uvicorn.run(self._app, host="0.0.0.0", port=self.port)
+
+        except Exception as err:
+            logger.exception("error while running app:", err)
```
danielgerlag, line 114: Any terminal error, we need to log to `/dev/termination-log`, in order for the error to show up when inspecting the drasi resources using the CLI. How does the logging config from `logger.py` get consumed in the context of using this library?

reactions/sdk/python/Makefile
```diff
@@ -0,0 +1,19 @@
...
+	poetry install --with dev
+	npm install --prefix ../../../typespec
+
+generate-types:
+	$(MAKE) install-dependencies
+	npm run clean --prefix ../../../typespec
+	npm run build ./view-svc --prefix ../../../typespec
+	npm run build ./query-output --prefix ../../../typespec
+	poetry run datamodel-codegen --input-file-type jsonschema --input ../../../typespec/query-output/_generated/@typespec/json-schema --output-model-type pydantic_v2.BaseModel --use-annotated --output ./drasi_reaction/models
+	poetry run datamodel-codegen --input-file-type jsonschema --input ../../../typespec/view-svc/_generated/@typespec/json-schema --output-model-type pydantic_v2.BaseModel --use-annotated --output ./drasi_reaction/models
```
danielgerlag, line 13: I don't think the sdk uses any of the `view-svc` models?

danielgerlag (commented):
Thank you for this contribution, this is awesome.  We can publish the library on official channels from our side.

reactions/sdk/python/Makefile
sjrusso8: Yes the sdk doesn't use them. I just included them for completeness. I can remove them

reactions/sdk/python/drasi_reaction/sdk.py
sjrusso8: Great question. Right now the logger is initialized when the function `config_logging` is called which is in the file `logger.py`. 

In that file there is a configuration dict defined so that any `ERROR` level logs to the file location `/dev/termination-log` as well as stout. All `INFO` level logs go to stout.

I did run into an issue when just running a dummy script to make sure the server would start that the logger would fail to initialize because it did not have the permission to log to `dev/`.

reactions/sdk/python/tests/unit/test_sdk.py
```diff
@@ -0,0 +1,54 @@
+from pathlib import Path
+from unittest.mock import Mock
+
+import pytest
+
+from drasi_reaction.sdk import DrasiReaction
```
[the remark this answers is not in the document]
sjrusso8: I can change it! It does modify the folder structure a little which is probably more extensible if the sdk evolves into more than just reactions. So it'll look like this now

```bash
drasi/
├── __init__.py
├── reaction/
│   ├── __init__.py
│   ├── sdk.py
│   ├── ....
│   ├── models/
...
```

reactions/sdk/python/drasi_reaction/sdk.py
sjrusso8: I see what you mean. It doesn't need *all* errors to be logged, but only the terminal error needs to be written to the file. I'll update accordingly

reactions/sdk/python/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+pip install drasi_reaction_sdk
+```
+
+### Simple example
+
+The following example logs the various parts of the incoming change event from a [Continuous Query](https://drasi.io/concepts/continuous-queries/).
+
+```python
```
ruokun-niu, line 17: nit: It might be nice to include an example where we are accessing the different types of results from the ChangeEvent. I think it's nice to be consistent with the examples that were provided in the other SDKs

sjrusso8: I agree. Updated!

ruokun-niu (approved):

danielgerlag (approved):

Issue #119, opened 2024-11-25
drasi install fails if dapr install isn't named "dapr"


markjgardner:
### Steps to reproduce

Install the managed extension on AKS and name the installation `dapr-ext`
`az k8s-extension create --cluster-type managedClusters --cluster-name mycluster-aks --resource-group myaks-rg --name **dapr-ext** --extension-type Microsoft.Dapr --auto-upgrade-minor-version false`

### Observed behavior

If you have installed dapr using a name other than `dapr` the drasi install fails because it doesn't recognize that dapr is already installed and then tries to install it again, colliding with existing resources:

```
$ drasi init -n drasi
Installing Drasi with version 0.1.6 from registry ghcr.io
ℹ Dapr not installed
✗ Error intalling Dapr: Unable to continue with install: ServiceAccount "dapr-injector" in namespace "dapr-system" exist
Error: Unable to continue with install: ServiceAccount "dapr-injector" in namespace "dapr-system" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "dapr": current value is "dapr-ext"
```

### Desired behavior

User should be able to indicate that dapr is already installed or specify the name used during install so that installation can be correctly detected by the drasi install process.

### Workaround

Reinstall dapr on the cluster, making sure to name it `dapr`

### Drasi Version

$ drasi version
Drasi CLI version: 0.1.6

### Operating system

Ubuntu 22.04

### Additional context

I believe this hardcoded label selector is the problem:
https://github.com/drasi-project/drasi-platform/blob/cd2f793d67468c2f8993682c5b140a022c0133b3/cli/service/installer.go#L625

### Would you like to support us?

- [X] Yes, I would like to support you

markjgardner:
As an alternative, maybe the label `app.kubernetes.io/part-of: "dapr"` would be a better option as it should be more consistent across clusters and different installs?
https://github.com/dapr/dapr/blob/master/charts/dapr/values.yaml#L85

markjgardner:
In testing this change there appears to be some really tight coupling between drasi and dapr (specifically 1.9.0). Can someone explain why that is before I go and try ripping all of that stuff out?

ruokun-niu:
Hi @markjgardner , just checking in to see if there's anything you need from us?

markjgardner:
[...]

PR #120, opened 2024-11-25
Add cli flag to enable skipping dapr install

markjgardner:
# Description

Allow users to skip installing dapr in situations where it is already installed on the cluster (and may not be detected by the built in check)

## Type of change
BugFix
Fixes: #119


agentofreality:
We have improved detection of pre existing installs of DAPR, so this may be sufficient to resolve the issues you were trying to resolve.

We are going to close this PR due to inactivity. Please raise a new PR if there is still an issue and you want to contribute a solution.

PR #121, opened 2024-11-27
Review Debezium reaction

ruokun-niu:
# Description

The Drasi Debezium Reaction connector generates a Debezium-compatible _data change event_ for each Added, Updated, or Removed in a Drasi result for a given Continuous Query.

The reaction takes the following configuration properties:
|Property|Description
|--- |---
|`queries`| The list of continuous queries you would like the Debezium Reaction to listen to and publish change events for.
|`properties.brokers`| The comma-separated list of Kafka brokers to write to, for example `test-kakfa:9092` which is the name of the server set up by applying the `test-kafka.yaml` file.
|`properties.topic`| The name of the Kafka topic to write to, for example `my-kafka-topic`.
|`properties.includeKey`| Whether to include the `key` in the resulting event. This defaults to `false` so only the value is included by default.
|`properties.includeSchemas`| Whether to include the `schema` in the resulting event. If `includeSchemas` is set to `true` this will also include the key schema, otherwise only the value schema is included. This defaults to `false`.

The exact format of the output from the Debezium reaction is explained in detail in the readme file. The skeleton JSON below shows the basic four parts of a standard change event:

```json
{
 "schema": { // schema for defining the structure of the key;  This will only be included if `properties.includeSchemas` is set to true
   ...
  },
 "payload": { // Event key
   ...
 },
 "schema": { // schema for defining the structure of the Continuous Query Result data
   ...
 },
 "payload": { // Continuous Query Result data
   ...
   ...
 },
}
```

## Type of change

Fixes: #issue_number


reactions/debezium/debezium-reaction/readme.md
```diff
@@ -1,363 +1,273 @@
...
+Note that unlike other connectors, the Drasi Debezium Reaction doesn't inherit type information from the underlying data sources (e.g. SQL, MongoDB, etc.). Instead, it infers the type information from the Continuous Query result JSON, representing the schema. This means that the type information in the event is not necessarily the same as the type information in the underlying data source, and are limited to the 6 broad JSON types:
+
+- `string`
+- `number`
+- `boolean`
+- `null`
+- `object`
+- `array`
+
+> ⚠️ This is one area where it potentially breaks compatibility with Debezium because the type information associated with the event schemas are expected to be Kafka Connect types, not JSON types. This will need to be addressed in the future.
```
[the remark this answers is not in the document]
ruokun-niu: Planning to discuss this during the sync today... we can either update all of the inconsistent data types from the change events to the debezium schema types, or leave them as they are

reactions/debezium/debezium-reaction/Services/DebeziumServices.cs
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+
+using Confluent.Kafka;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Drasi.Reactions.Debezium.Services
+{
+	public class DebeziumService
```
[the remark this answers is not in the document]
ruokun-niu: Hmm... I originally wanted to store the kafka connection instance in this class, but ended up putting it somewhere else. I'll investigate to see if we can remove this class

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -0,0 +1,308 @@
...
+
+		using var producer = new ProducerBuilder<Null, string>(_debeziumService.Config).Build();
+		await ProcessResults(producer, metadata, evt.AddedResults, "c");
+
+		var updatedResults = new List<Dictionary<string, object>>();
+		for (int i = 0; i < evt.UpdatedResults.Length; i++)
+		{
+			var currResultDict = new Dictionary<string, object>
+			{
+				["before"] = evt.UpdatedResults[i].Before,
```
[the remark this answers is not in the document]
ruokun-niu: Good point

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+		
+		// A list of brokers; represented as a comma-separated string
+		string brokers = config.GetValue<string>("brokers") ?? throw new ArgumentNullException("Debezium brokers is required");
+
+		ProducerConfig producerConfig = new ProducerConfig
+		{
+			BootstrapServers = brokers
+		};
+
+		_producer = new ProducerBuilder<Null, string>(producerConfig).Build();
```
danielgerlag, line 54: The Change Handler is a scoped service, meaning a new instance of the service is created for each request, so constructing the producer here will still create a new kafka connection for every message. It's best to take the `IProducer` as a constructor parameter and register it on the IoC container as a singleton.  I also makes this class more testable because you could pass a mock object in place of this.

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+		var metadata = new EventMetadata(evt);
+
+		_logger.LogInformation("Processing {QueryId}", metadata.QueryId);
+
+		await ProcessResults(metadata, evt.AddedResults, "c");
+
+		var updatedResults = new List<Dictionary<string, object>>();
+		for (int i = 0; i < evt.UpdatedResults.Length; i++)
+		{
+			var currResultDict = new Dictionary<string, object>
```
danielgerlag, line 70: Seems like this bit should be part of the formatter?

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+			string eventString = _formatter.FormatEvent(metadata, res, op);
+
+			_logger.LogInformation($"dataChangeEvent: {eventString}");
+			try
+			{
+				var deliveryReport = await _producer.ProduceAsync(_topic, new Message<Null, string> { Value = eventString }, CancellationToken.None);
+
+				if (deliveryReport.Status != PersistenceStatus.Persisted)
+				{
+					_logger.LogInformation($"Delivery failed: {deliveryReport.Message.Value}");
```
danielgerlag, line 97: If the delivery failed, should we throw an exception, so that the message from the dapr queue in not acknowledged, and it can retry?

reactions/debezium/debezium-reaction/Services/DataChangeEventFormatter.cs
```diff
@@ -0,0 +1,265 @@
...
+    };
+
+    public DataChangeEventFormatter(bool includeKey, bool includeSchemas)
+    {
+        _includeKey = includeKey;
+        _includeSchemas = includeSchemas;
+
+    }
+
+    public string FormatEvent(EventMetadata metadata, Dictionary<string, object> res, string op)
```
danielgerlag, line 36: It might make more sense for this to take in a `ChangeEvent` and return a list of "kakfa messages"

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
ruokun-niu: Yep, given that we want to modify the `FormatEvent` function to accept a `ChangeEvent`, we can just put this logic in the formatter

reactions/debezium/debezium-reaction/Program.cs
```diff
@@ -1,13 +1,43 @@
...
 var reaction = new ReactionBuilder()
             .UseChangeEventHandler<DebeziumChangeHandler>()
             .ConfigureServices((services) =>
-            {})
+            {
+                services.AddSingleton<IProducer<Null, string>>(sp =>
+                {
+                    var config = sp.GetRequiredService<IConfiguration>();
+                    // A list of brokers; represented as a comma-separated string
+                    string brokers = config.GetValue<string>("brokers") ?? throw new ArgumentNullException("Debezium brokers is required");
```
danielgerlag, line 31: Are they `Debezium` brokers or `Kafka` brokers?

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -29,31 +29,20 @@ class DebeziumChangeHandler : IChangeEventHandler
...
-		{
-			BootstrapServers = brokers
-		};
-
-		_producer = new ProducerBuilder<Null, string>(producerConfig).Build();
+		
+		_producer = producer;
 
-		_formatter = new DataChangeEventFormatter(_includeKey, _includeSchemas);
+		_formatter = new DataChangeEventFormatter();
```
danielgerlag, line 45: This could also be injected from the IoC container

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
```diff
@@ -62,51 +51,30 @@ public async Task HandleChange(ChangeEvent evt, object? queryConfig)
...
+		foreach (string message in messageList)
 		{
-			string eventString = _formatter.FormatEvent(metadata, res, op);
-
-			_logger.LogInformation($"dataChangeEvent: {eventString}");
+			_logger.LogInformation($"dataChangeEvent: {message}");
 			try
 			{
-				var deliveryReport = await _producer.ProduceAsync(_topic, new Message<Null, string> { Value = eventString }, CancellationToken.None);
+				var deliveryReport = await _producer.ProduceAsync(_topic, new Message<Null, string> { Value = message }, CancellationToken.None);
```
danielgerlag, line 61: Does Kafka support sending a group of messages as a transaction?
If so, it might be a good idea to send the list of messages within a transaction.  That way, if message 5 fails and we end up resending the whole batch, there won't be duplicates.

reactions/debezium/debezium-reaction/Program.cs
ruokun-niu: kafka brokers

reactions/debezium/debezium-reaction/Services/ChangeHandler.cs
ruokun-niu: I think we can achieve this through this https://docs.confluent.io/platform/current/clients/confluent-kafka-dotnet/_site/api/Confluent.Kafka.IProducer-2.html#Confluent_Kafka_IProducer_2_BeginTransaction

ruokun-niu: We will commit the transaction if no exceptions were thrown, otherwise we will abort it

danielgerlag (approved):

agentofreality (approved):

PR #123, opened 2024-12-03
VS Code extension config update and gh workflow

ruokun-niu:
# Description

- Added a workflow with the name `Build and Publish Visual Studio Code Extension`. This workflow can be triggered manually(requires a PAT; see https://code.visualstudio.com/api/working-with-extensions/publishing-extension#get-a-personal-access-token). It accepts an input called `version`, and it will download the vsix from this release and publish it to the marketplace.
- Added an icon for the extension
- configured the publisher, version and description in the `package.json` file

## Type of change

Fixes: #issue_number


danielgerlag (commented):
I think we should release this extension in band with the management API (drasi platform) and have them share version numbers. That way, you know if you have drasi 0.1.5, that the 0.1.5 extension will work with it.

ruokun-niu:
> I think we should release this extension in band with the management API (drasi platform) and have them share version numbers. That way, you know if you have drasi 0.1.5, that the 0.1.5 extension will work with it.

Good point. I'll update this

ruokun-niu:
> I think we should release this extension in band with the management API (drasi platform) and have them share version numbers. That way, you know if you have drasi 0.1.5, that the 0.1.5 extension will work with it.

@danielgerlag I have modified the workflow based on our discussion today during the sync. The workflow will download the extension from a release and publish it from there

danielgerlag (approved):

agentofreality (approved):

PR #129, opened 2024-12-12
Updated the release workflow to use makefiles and updated rust version in lint workflow

ruokun-niu:
# Description

- The release workflow currently build and push multi-arch images using the Makefiles
- Set rust version to 1.83.0 in the lint workflow; this is consistent with `drasi-core`
- Updated the dockerfile of the Signalr Reaction to fix build error (This issue came up when we tried to build arm64 images in the github workflow runner)

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -100,7 +100,7 @@ jobs:
         run: |
           cd ${{ matrix.component.path }}
           DOCKER_TAG_VERSION=${{ inputs.tag }} \
-          IMAGE_PREFIX="ghcr.io/ruokun-niu" \
+          IMAGE_PREFIX="ghcr.io/drasi-project" \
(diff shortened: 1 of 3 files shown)


.github/workflows/draft-release.yml
```diff
@@ -131,30 +142,23 @@ jobs:
...
+          path: /tmp/.buildx-cache
+          key: buildx-${{ matrix.component.name }}
+          restore-keys: |
+            buildx-${{ matrix.component.name }}
+      
+      - name: Build and Push to GHCR
+        run: |
+          cd ${{ matrix.component.path }}
+          DOCKER_TAG_VERSION=${{ inputs.tag }} \
+          IMAGE_PREFIX="ghcr.io/drasi-project" \
```
danielgerlag, line 158: Could we put the `IMAGE_PREFIX` in a variable?

agentofreality (approved):

.github/workflows/draft-release.yml
ruokun-niu: I have updated this the workflow to accept an addtional input called `image_prefix`. This input is not required and it will use 'ghcr.io/drasi-project' if not set

PR #136, opened 2025-01-03
Source Subscription deletion

ruokun-niu:
# Description

Currently, when a continuous query is deleted, the subscription info in the change router is still persisted. This PR focuses on the deletion of subscription when a continuous query is deleted. Specifically, this is achieved by the following steps:
- When a `DELETE` command is received in the query worker, the worker will iterate through all of the sources that the query is subscribed to, and then invoke the `unsubscription` method in the sources' query-api (this is achieved through dapr service invocation).
- When the `unsubscription` method is invoked, the query api will publish the following message to the Source's change router:
```json
{"payload":{"queryId":"<query-id>","queryNodeId":"<query-node-id>","source":{"db":"Drasi","table":"SourceUnsubscription"}}}
```
This message has a similar format as the SourceSubscription message
- When the change router receives this message, it will format the state entry key and delete it from the dapr state store

## Type of change

Fixes: #issue_number


query-container/query-host/src/source_client.rs
```diff
@@ -84,6 +84,45 @@ impl SourceClient {
...
+    ) -> Result<(), UnsubscriptionError> {
+        let app_id = format!("{}-query-api", subscription_id);
+        let data = json!({
+            "queryNodeId": query_container_id,
+            "queryId": query_id,
+        });
+        
+        let resp = match self
+            .client
+            .post(format!("http://{}/unsubscription", app_id))
```
danielgerlag, line 102: To remain "RESTful", I think this should be a `DELETE` verb on the `subscription/{queryNodeId}/{queryId}` route.

sources/shared/query-api/src/main.rs
```diff
@@ -186,6 +207,45 @@ async fn dispatch_control_event(
     Ok(())
 }
 
+async fn dispatch_unsubscription_event(
+    unsubscription_request: &UnsubscriptionRequest,
+    state: &Arc<AppState>,
+) -> Result<(), axum::http::Response<axum::body::Body>> {
+    let publisher = &state.publisher;
+    let unsubscription_event = api::UnsubscriptionEvent {
```
danielgerlag, line 215: Could this just be a `ControlEvent` with an `op` of `d` instead of `i` on the `SourceSubscription` table?

sources/shared/change-router/src/main.rs
```diff
@@ -391,6 +391,31 @@ async fn process_changes(
                     } else {
                         // TODO - supprt other ops on SourceSubscriptions
                     }
+                } else if change["payload"]["source"]["table"] == "SourceUnsubscription" {
```
danielgerlag, line 394: Could we refactor these deep nested if else statements into a `match` statement?

danielgerlag (changes requested):

sources/shared/query-api/src/main.rs
ruokun-niu: Good point. Will modify this

sources/shared/query-api/src/main.rs
```diff
@@ -186,6 +204,54 @@ async fn dispatch_control_event(
...
+    let control_event = ControlEvent {
+        op: "d".to_string(),
+        ts_ms: Utc::now().timestamp_millis() as u64,
+        payload: SubscriptionPayload {
+            source: Source {
+                db: "Drasi".to_string(),
+                table: "SourceSubscription".to_string(),
+            },
+            before: None,
+            after: request.clone(),
```
danielgerlag, line 228: typically for these debezium style payloads, the properties for a "delete" will be in the "before" field.

danielgerlag (approved):

agentofreality (approved):

PR #138, opened 2025-01-07
Configured the EventGrid Reaction to use different schemas

ruokun-niu:
# Description

Currently, the EventGrid reaction only supports eventgrid topics that use `CloudEvents` as the schema. 

This PR configures the Reaction by adding a new property called `eventGridSchema`. The default value of this property is set to `CloudEvents`, which publishes CloudEvents to the EventGrid topic. This property can be set to `EventGrid` to publish events in the EventGrid schema to the topic.

## Type of change

Fixes: #issue_number


cli/service/resources/default-reaction-providers.yaml
```diff
@@ -66,6 +66,9 @@ spec:
         type: string
       eventGridKey:
         type: string
+      eventGridSchema:
+        type: string
+        default: CloudEvents
```
danielgerlag, line 71:
```suggestion
      eventGridSchema:
        type: string
        enum:
          - "CloudEvents"
          - "EventGrid"
        default: "CloudEvents"
```

This will also enable validation of one of these 2 options.

reactions/azure/eventgrid-reaction/Services/ChangeHandler.cs
```diff
@@ -47,27 +54,53 @@ public async Task HandleChange(ChangeEvent evt, object? queryConfig)
         switch(_format)
         {
             case OutputFormat.Packed:
-                CloudEvent egEvent = new CloudEvent(evt.QueryId, "Drasi.ChangeEvent", _formatter.Format(evt));
-                var resp = await _publisherClient.SendEventAsync(egEvent);
-                if (resp.IsError) 
-                {
-                    _logger.LogError($"Error sending message to Event Grid: {resp.Content.ToString()}");
-                    throw new Exception($"Error sending message to Event Grid: {resp.Content.ToString()}");
+                if (_eventGridSchema == "cloudevents") {
```
danielgerlag, line 57: We could store an enum in `_eventGridSchema` to avoid doing a string comparison for every event.

cli/service/resources/default-reaction-providers.yaml
danielgerlag: Also, it looks like we are missing the format (packed/unpacked) config value?

ruokun-niu: good catch

ruokun-niu: I think maybe in the future we should make this `format` property a default value in the reaction_provider, so that we dont have to implement this for every reaction

agentofreality (approved):

amansinghoriginal (approved):

PR #147, opened 2025-01-31
Remove Fixed Dapr Version Constraint

ruokun-niu:
# Description

We previously had to use a fixed verison of the dapr runtime (1.10.0), as we would run into the following error when we try to run `drasi init` with the latest version of dapr:
```bash
error saving actor transaction state: using transactions with MongoDB requires connecting to a replica set
```
This also restricted us to use the 1.9.0 version of the dapr sidecar.

Initially, I thought this was an issue with how we setup mongo in the Drasi cluster. However, after exploring a bit, I realized that the issue was with how we set up mongo as the dapr statestore. When we specify the mongo during the state store definition, we need to specifically point towards a mongo replicaset, which can be defined in the `connectionString` field.

NOTE: In the [official dapr documentation](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-mongodb/) on using mongo as a state store, `connectionString` is not listed as a valid field. However, according to this [discord discussion](https://discord.com/channels/778680217417809931/901141713089863710/1274079252190330952) and this [PR](https://github.com/dapr/components-contrib/pull/2659), this field is actually added in 2024 and they just forgot to update the doc site. By defining a connection string to the replicaset in `infra.yaml`, dapr will correctly use the replica set as the mongo actor state store, which allows us to install Drasi with the latest version of dapr without any trouble.

**The description below is generated by copilot :)**

This pull request includes changes to the `cli` package to update the configuration for Dapr versions and MongoDB connection settings. The most important changes are as follows:

### Updates to Dapr version configuration:

* [`cli/cmd/init.go`](diffhunk://#diff-1d8fe5889cb8e0e5a24f1fe9f788c4f2b1f843be25c23075fda78f87d2c1502fL111-R112): Updated the default values for `dapr-runtime-version` to be empty (which is the equivalent of latest) and `dapr-sidecar-version` to "latest".

### Changes to MongoDB connection settings:

* [`cli/service/resources/infra.yaml`](diffhunk://#diff-26c5e37903160af44d120cd991debd1e31837d5d5ae67b8073d769bb938d3e38L319-R324): Replaced individual host, username, and password metadata with a single `connectionString` for MongoDB, including a comment explaining the necessity of the `replicaSet` parameter.
* [`cli/service/resources/infra.yaml`](diffhunk://#diff-26c5e37903160af44d120cd991debd1e31837d5d5ae67b8073d769bb938d3e38L347): Removed an unnecessary blank line.

## Type of change

Fixes: #issue_number


.github/workflows/build-test.yml
```diff
@@ -372,7 +372,7 @@ jobs:
     permissions:
       packages: write
       contents: read
-    runs-on: e2e-tester
+    runs-on: ubuntu-latest
```
danielgerlag, line 375: Don't forget about this.

cli/service/resources/infra.yaml
```diff
@@ -316,12 +316,12 @@ spec:
...
   version: v1
   metadata:
-    - name: host
-      value: drasi-mongo:27017
-    - name: username
-      value:
-    - name: password
-      value:
+    - name: connectionString
+      value: "mongodb://drasi-mongo-0.drasi-mongo:27017/?replicaSet=rs0"
```
danielgerlag, line 320: Is this always going to connect to the `-0` instance? Is there somekind of service proxy to route to any of the available instances?

ruokun-niu: I think we can use the service name (`drasi-mongo`) instead. I'll test this to make sure things are working

amansinghoriginal (commented):
DAPR version change looks good to me. But there seems to be some unrelated changes in this PR also. Is my understanding correct?

reactions/platform/result-reaction/Program.cs
```diff
@@ -31,7 +31,7 @@
 	queryIds.Add(queryId);
 }
 
-
+Console.WriteLine("QueryContainerId: " + queryContainerId);
```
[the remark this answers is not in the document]
ruokun-niu: Good catch. Thanks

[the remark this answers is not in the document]
ruokun-niu: The addition of `await` on line 142 in this file is minor fix for the result reaction. Encountered this occasionally when I attempted the smoke test

cli/service/resources/infra.yaml
ruokun-niu: I have just tested this. We can use the service name in the connectionstring  (`"mongodb://drasi-mongo:27017/?replicaSet=rs0")

amansinghoriginal (approved):

agentofreality (approved):

PR #149, opened 2025-02-05
Gremlin Reaction hotfix

ruokun-niu:
# Description

Improved the following areas in the Gremlin Reaction:
- If the same parameter name is used multiple times in a single Gremlin command (for example `g.addV('Item').property('ItemName', @Name).property('Name',@Name)`), an exception will be thrown. I have added a check to catch this
- When a certain type of result is received (e.g. DeletedResults), even if the user has not supplied a Gremlin command for that type, the Gremlin client will still attempt to execute the null command
- Added a test for this scenario

## Type of change

Fixes: #issue_number

### DIFF
--- a/reactions/gremlin/gremlin-reaction/Dockerfile
+++ b/reactions/gremlin/gremlin-reaction/Dockerfile
@@ -15,14 +15,14 @@
 
 #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
 
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-cbl-mariner2.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0-cbl-mariner2.0 AS build
 ARG BUILD_CONFIGURATION=Release
 WORKDIR /src
 COPY . . 
 RUN dotnet restore
 RUN dotnet build -c $BUILD_CONFIGURATION -o /app/build
 
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/aspnet:8.0-cbl-mariner2.0 AS final
+FROM mcr.microsoft.com/dotnet/aspnet:8.0-cbl-mariner2.0 AS final
 WORKDIR /app
 COPY --from=build /app/build .
 ENTRYPOINT ["/app/gremlin-reaction"]
--- a/reactions/gremlin/gremlin-reaction/Services/GremlinService.cs
+++ b/reactions/gremlin/gremlin-reaction/Services/GremlinService.cs
@@ -138,6 +138,11 @@ namespace Drasi.Reactions.Gremlin.Services {
 
         public void ProcessAddedQueryResults(Dictionary<string, object> res)
         {
+            if (_addedResultCommand == null)
+            {
+                _logger.LogInformation("No Deleted Result Command Specified");
+                return;
+            }
             string newCmd = _addedResultCommand;
 
             // Dictionary to hold the parameters for the command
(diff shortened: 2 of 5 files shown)


.github/workflows/build-test.yml
```diff
@@ -372,7 +372,7 @@ jobs:
     permissions:
       packages: write
       contents: read
-    runs-on: e2e-tester
+    runs-on: ubuntu-latest
```
danielgerlag, line 375: The e2e-tester has more resources, so tests should run faster, more reliable

ruokun-niu: Yeah our test runner was deleted. I just wanted to see if the tests were passing for now. 

We should discuss for alternatives during tomorrow's sync

reactions/gremlin/gremlin-reaction/Dockerfile
```diff
@@ -15,14 +15,14 @@
 
 #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
 
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0-cbl-mariner2.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0-cbl-mariner2.0 AS build
```
danielgerlag, line 18: I think the multi arch release needs `--platform=$BUILDPLATFORM`

reactions/gremlin/gremlin-reaction/Services/GremlinService.cs
```diff
@@ -138,6 +138,11 @@ public GremlinService(IConfiguration configuration, ILogger<GremlinService> logg
 
         public void ProcessAddedQueryResults(Dictionary<string, object> res)
         {
+            if (_addedResultCommand == null)
+            {
+                _logger.LogInformation("No Deleted Result Command Specified");
```
danielgerlag, line 143: This is the added command or deleted command?

reactions/gremlin/gremlin-reaction/Services/GremlinService.cs
```diff
@@ -168,6 +178,11 @@ public void ProcessAddedQueryResults(Dictionary<string, object> res)
 
         public void ProcessUpdatedQueryResults(UpdatedResultElement updatedResult)
         {
+            if (_updatedResultCommand == null)
+            {
+                _logger.LogInformation("No Deleted Result Command Specified");
```
danielgerlag, line 183: This is the updated command or deleted command?

reactions/gremlin/gremlin-reaction/Dockerfile
ruokun-niu: Oh yep. Although I think for the second `FROM` we should use `--platform=$TARGETPLATFORM` instead of `BUILDPLATFORM`. The `BUILDPLATFORM` will always be the same arch as the host machine. I'll update this 


https://stackoverflow.com/questions/73978929/how-to-build-docker-image-for-multiple-platforms-with-cross-compile

reactions/gremlin/gremlin-reaction/Services/GremlinService.cs
ruokun-niu: good catch

ruokun-niu:
Based on today's dicussion during the sync, I have updated the Gremlin reaction in the following areas:
- In the constructor of `GremlinService`, we retrieve a list of parameters for each Gremlin command using regex and add them to a respective list (stored as class variables). During this step, we will check to see if there are any duplicate parameters, and we will skip the duplicates. At this step, we will also update the Gremlin command by removing the `@` from the parameters. We will store the updated commands as class variables, removing the need to update them on every change event.
- When we are handling the change events, we will populate a dictionary where its keys are the parameters and the values are coming from query result. This dictionary will be used as the parameters map when we submit the request to gremlin (see https://tinkerpop.apache.org/docs/current/reference/#parameterized-scripts for more details). We will also do a comparsion to see if the number of entires in the map matches the number of parameters in the parameter list

danielgerlag (approved):

PR #152, opened 2025-02-12
EventHub Source

danielgerlag:
# EventHub Source

- Refactored the EventHub source to use the .net SDK
- Added support for Managed Identities


cli/service/resources/default-source-providers.yaml
```diff
@@ -250,6 +250,8 @@ spec:
     properties:
       connectionString:
         type: string
+      host:
```
ruokun-niu, line 253: Is the `host` field only used (and required?) if we are using managed identities?

ruokun-niu (approved):

amansinghoriginal (approved):

PR #155, opened 2025-02-19
Review debug reaction

ruokun-niu:
# Description

- Use React instead of Blazor to build the frontend
- Updated the code to use the Reaction SDK
- Updated the raw event stream to update automatically when a new event is received
- Updated the reaction provider with frontend and backend pod configs

## Type of change

Fixes: #issue_number


reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Text;
+using System.Net.WebSockets;
+
+var builder = WebApplication.CreateBuilder(args);
+
+var configuration = BuildConfiguration();
+
+var pubsubName = configuration.GetValue<string>("PubsubName", "drasi-pubsub");
+var configDirectory = configuration.GetValue<string>("QueryConfigPath", "/etc/queries");
+var queryContainerId = configuration.GetValue<string>("QueryContainer", "default");
```
danielgerlag, line 33: How can this work across any query container?

danielgerlag (changes requested):

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,186 @@
...
+		private readonly DaprClient _daprClient;
+
+
+		private readonly ConcurrentDictionary<string, QueryResult> _results = new();
+
+		private readonly LinkedList<JsonElement> _rawEvents = new();
+		private readonly string _queryDir;
+		private readonly string _queryContainerId;
+
+		private readonly WebSocketService _webSocketService;
```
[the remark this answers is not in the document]
ruokun-niu: Good point. thanks

reactions/sdk/dotnet/Drasi.Reaction.SDK/Reaction.cs
```diff
@@ -54,6 +54,8 @@ public void Dispose()
         
     }
 
+    public WebApplication App() => _app;
```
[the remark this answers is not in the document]
ruokun-niu: I'll take a look at the signalR reaction. Thanks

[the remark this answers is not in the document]
danielgerlag: It also exposes dapr endpoints to the outside world.

reactions/platform/debug-reaction/debug-reaction.Server/Services/IQueryDebugService.cs
```diff
@@ -12,20 +12,22 @@
...
\ No newline at end of file
+using Drasi.Reactions.Debug.Server.Models;
+using System.Text.Json;
+
+namespace Drasi.Reactions.Debug.Server.Services
+{
+	public interface IQueryDebugService : IHostedService
+	{
+		Task<QueryResult> GetQueryResult(string queryId);
+		Task ProcessRawChange(string queryId, JsonElement change);
```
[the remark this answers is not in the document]
ruokun-niu: Yep. Good point

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
ruokun-niu: Just clarifying this: are you asking how the Reaction will handle situations where the queries are hosted in different query containers?

danielgerlag: yes

ruokun-niu: I think we can use something like https://github.com/drasi-project/drasi-platform/blob/2227a1c96e5e9bd910af3a8afad73fbcaa53edb5/reactions/sdk/dotnet/Drasi.Reaction.SDK/Services/ManagementClient.cs#L20 to retrieve the query container IDs for the continuous queries. This allows us to handle situations like this

reactions/platform/debug-reaction/debug-reaction.Server/Controllers/QueryController.cs
```diff
@@ -0,0 +1,73 @@
...
+// 	return Results.Json(result);
+// });
+
+namespace Drasi.Reactions.Debug.Server.Controllers;
+using Microsoft.AspNetCore.Mvc;
+using Drasi.Reactions.Debug.Server.Services;
+
+[ApiController]
+[Route("query")]
+public class QueryController : ControllerBase
```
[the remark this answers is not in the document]
ruokun-niu: Planning to do the followings:

`queries`
- used for retrieving a list of queries that the reaction is subscribed to

`queries/{queryId}`
- Websocket for the query results

`queries/{queryId}/current-results`
- for retrieving the current result set of a query. This is used when the page loads up and when the user clicks the `reinitialize query cache` button

`queries/{queryId}/debug-information`
- for debug info like query host, status and etc.

reactions/platform/debug-reaction/debug-reaction.Server/Controllers/QueryController.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    [HttpGet("initialize/{queryId}")]
+    public async Task<IActionResult> InitializeQuery(string queryId)
+    {
+        var result = await _debugService.GetQueryResult(queryId);
+        return Ok(result);
+    }
+
+    [HttpGet("reinitialize/{queryId}")]
```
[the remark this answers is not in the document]
ruokun-niu: I have added a short description of the API in the comment above

reactions/platform/debug-reaction/debug-reaction.Server/Services/IQueryDebugService.cs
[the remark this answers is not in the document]
ruokun-niu: For the raw change events that are used in the event stream, they are still stored in json as they will be rendered into json objects. 

For the methods that are required for processing the change events and control events, I have updated them to use `ChangeEvent` and `ControlEvent` respectively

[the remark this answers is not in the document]
ruokun-niu: However I think it still makes sense to store the results of a query as a `JsonElement`. this makes it easier for us to store results for different queries, which will have different fields

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+					context.Response.StatusCode = 400; // Bad Request if not a WebSocket request
+				}
+			}
+			else
+			{
+				await next();
+			}
+		});
+
+		server.Urls.Add("http://0.0.0.0:5195");
```
danielgerlag, line 159: Should we get the port from an env variable?

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+		await Task.WhenAll(reaction.StartAsync(ShutdownToken.Token),server.RunAsync());
+	}
+	public static TaskCompletionSource StartupTask { get; } = new();
+
+	public static CancellationTokenSource ShutdownToken { get; } = new();
+}
+
+
+
+public class ChangeHandler : IChangeEventHandler
```
danielgerlag, line 183: For .NET, the convention is generally to put a class per file.

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+		_debugService = debugService;
+		_logger = logger;
+	}
+
+	public async Task HandleChange(ChangeEvent evt, object? queryConfig)
+	{
+		var queryId = evt.QueryId;
+		var jsonEvent = JsonSerializer.Deserialize<JsonElement>(evt.ToJson());
+		await _debugService.ProcessRawEvent(jsonEvent);
+		await _debugService.ProcessRawChange(evt);
```
danielgerlag, line 200: Why does it process the same thing twice for in two different formats?

reactions/platform/debug-reaction/debug-reaction.Server/Services/IChangeBroadcaster.cs
```diff
@@ -12,13 +12,17 @@
...
+using Drasi.Reactions.Debug.Server.Models;
+
+namespace Drasi.Reactions.Debug.Server.Services
+{
+    public interface IChangeBroadcaster
+    {
+        void AddConnection(string socketName, WebSocket webSocket);
+        Task BroadcastToQueryId(string socketName, QueryResult message);
+
+        Task BroadcastToStream(string socketName, LinkedList<JsonElement> message);
```
danielgerlag, line 26: What is the difference between `BroadcastToStream` and `BroadcastToQueryId`?

reactions/platform/debug-reaction/debug-reaction.Server/Services/IQueryDebugService.cs
```diff
@@ -12,20 +12,21 @@
...
+	public interface IQueryDebugService : IHostedService
+	{
+		Task<QueryResult> GetQueryResult(string queryId);
+		Task ProcessRawChange(ChangeEvent change);
+		Task ProcessControlSignal(ControlEvent change);
+		IEnumerable<string> ActiveQueries { get; }
+		Task<Dictionary<string, object>> GetDebugInfo(string queryId);
+		Task<LinkedList<JsonElement>> GetRawEvents();
+
+		Task ProcessRawEvent(JsonElement change);
```
danielgerlag, line 30: What is the difference between `ProcessRawEvent` and `ProcessRawChange` + `ProcessControlSignal`?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+
+		public async Task ProcessRawEvent(JsonElement change)
+		{
+			lock (_rawEvents)
+			{
+				_rawEvents.AddFirst(change);
+				while (_rawEvents.Count > 100)
+					_rawEvents.RemoveLast();
+			}
+			await _webSocketService.BroadcastToStream("stream", _rawEvents);
```
danielgerlag, line 71: Do we broadcast the new event plus the previous 99 every time there is a new event?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+		{
+			return _rawEvents;
+		}
+
+		public async Task ProcessRawEvent(JsonElement change)
+		{
+			lock (_rawEvents)
+			{
+				_rawEvents.AddFirst(change);
+				while (_rawEvents.Count > 100)
```
danielgerlag, line 68: Why do we keep 100?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+		public async Task ProcessRawChange(ChangeEvent change)
+		{
+			var queryId = change.QueryId;
+			if (!_results.ContainsKey(queryId))
+				return;
+
+			var queryResult = _results[queryId];
+			foreach (var item in change.DeletedResults)
+			{
+				var result = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(item));
```
danielgerlag, line 106: Do we need to serialize and then deserialize?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+			}
+			foreach (var item in change.UpdatedResults)
+			{
+				var groupingKeys = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(item.GroupingKeys));
+				var before = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(item.Before));
+				var after = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.Serialize(item.After));
+				queryResult.Update(before, after, groupingKeys);
+			}
+
+			await _webSocketService.BroadcastToQueryId(queryId, queryResult);
```
danielgerlag, line 122: Do we send the entire result set for every incremental change that comes in?
Could we just send the incremental change?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+				case ControlSignalKind.BootstrapStarted:
+					queryResult.Clear();
+					break;
+			}
+
+			await _webSocketService.BroadcastToQueryId(queryId, queryResult);
+		}
+
+
+		private async Task<QueryResult> InitResult(string queryId)
```
danielgerlag, line 147: What happens if this method is invoked for the same query twice, before the first one completes?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+			}
+			catch (Exception ex)
+			{
+				result.Errors.Add("Error fetching initial data: " + ex.Message);
+				_logger.LogError(ex, "Error fetching initial data: " + ex.Message);
+			}
+
+			return result;
+		}
+		protected override async Task ExecuteAsync(CancellationToken stoppingToken)
```
danielgerlag, line 179: This doesn't seem to do anything? Do we need it?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
```diff
@@ -0,0 +1,201 @@
...
+namespace Drasi.Reactions.Debug.Server.Services
+{
+	public class QueryDebugService : BackgroundService, IQueryDebugService
+	{
+		private readonly IResultViewClient _queryApi;
+		private readonly IActorProxyFactory _actorProxyFactory;
+		private readonly DaprClient _daprClient;
+
+		private readonly ILogger<QueryDebugService> _logger;
+		private readonly ConcurrentDictionary<string, QueryResult> _results = new();
```
danielgerlag, line 38: Could we move this state to the front end, and just have the back end be a proxy for changes?

reactions/platform/debug-reaction/debug-reaction.client/src/pages/EventStream.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+import React, { useState, useEffect } from "react";
+
+function EventStream() {
+    const [stream, setStream] = useState([]);
+    const WEBSOCKET_URL = "ws://localhost:5195/ws/stream";
+
+    const fetchStream = async () => {
+        try {
+            const url = "http://localhost:5195/stream";
```
danielgerlag, line 25: If the front end and back end run in separate pods, how would `localhost` reach the backend?

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+		server.MapControllers();
+		server.UseWebSockets();
+
+
+		server.MapGet("/stream", async (IQueryDebugService debugService) =>
+		{
+			return Results.Json(await debugService.GetRawEvents());
+		});
+
+		server.Use(async (context, next) =>
```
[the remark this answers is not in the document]
ruokun-niu: Good point

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+public class Program
+{
+	public static async Task Main(string[] args)
+	{
+		var builder = WebApplication.CreateBuilder();
+        builder.Services.AddSingleton<IChangeBroadcaster, WebSocketService>();
```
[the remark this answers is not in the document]
ruokun-niu: Sorry good catch

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
ruokun-niu: I think one is for formatting the the events in the event stream. We can combine them into one

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: I think this limit was set in the previous implementation of the debug reaction. Should we use a new value?

reactions/platform/debug-reaction/debug-reaction.Server/Services/IChangeBroadcaster.cs
ruokun-niu: `BroadcastToStream` is used for sending updates to the websocket that is responsible for the raw event stream

reactions/platform/debug-reaction/debug-reaction.client/src/pages/EventStream.jsx
ruokun-niu: Currently we are still relying on port-forwarding both the frontend and the backend pods. I haven't looked into intra-cluster communication using k8s services yet, but I'll spend some time during this iteration

reactions/platform/debug-reaction/debug-reaction.Server/Services/IQueryDebugService.cs
ruokun-niu: `ProcessRawEvent` is used for sending event jsons to the websocket for processing raw event stream. This function also checks to see if the size of the stream reaches the capacity (100 items). If we move the state that stores the raw events to the frontend, I think we can remove the function `ProcessRawEvent` and just invoke the broadcast function directly

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: Will update this to send incremental updates

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: updated

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: updated

reactions/platform/debug-reaction/debug-reaction.Server/Program.cs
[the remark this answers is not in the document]
ruokun-niu: moved this to a controller

reactions/platform/debug-reaction/debug-reaction.client/src/pages/EventStream.jsx
danielgerlag: You could serve the front-end app content from the `/` route of the backend, and it could connect to the `/api/*` routes for the same address. That way, we don't need two containers.

ruokun-niu: Let me give it a try

ruokun-niu: Was able to get it working! Just need to clean up a few things

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: Good question. I did some searches online and it seems like one promising approach is to store/cache Task<QueryResult> using a `ConcurrentDicitionary`. That way concurrent invocation to this method will share the same result. Any thoughts on this approach @danielgerlag ?

reactions/platform/debug-reaction/debug-reaction.Server/Services/QueryDebugService.cs
ruokun-niu: I haveupdated this

reactions/platform/debug-reaction/Services/QueryDebugService.cs
```diff
@@ -12,154 +12,177 @@
...
+{
+	public class QueryDebugService : IQueryDebugService
+	{
+		private readonly IResultViewClient _queryApi;
+		private readonly IActorProxyFactory _actorProxyFactory;
+		private readonly DaprClient _daprClient;
+
+		private readonly ILogger<QueryDebugService> _logger;
+
+		private readonly ConcurrentDictionary<string, Task<QueryResult>> _activeQueries = new();
```
danielgerlag, line 39: If we are moving the query result state to the client, why do we need it keep it here too?

reactions/platform/debug-reaction/Services/QueryDebugService.cs
```diff
@@ -12,154 +12,177 @@
...
+
+		private readonly LinkedList<JsonElement> _rawEvents = new();
```
danielgerlag, line 41: Do we need to keep this data here?

ruokun-niu: Removed

reactions/platform/debug-reaction/Services/QueryDebugService.cs
ruokun-niu: I implemented this dictionary for storing the task of initilializing the query results. This dictionary is only used in the `InitResult` method, which uses the ResultViewClient to get the initial result for a query.

Let us say that we have a query called `query1` that is trying to initialize. In the first invocation, the method will check the dictionary. `query1` is not present in the dict, it will execute lambda function and store the Task<QueryResult> in the dictionary. If there is a second invocation (before the first one completes), it will wait and return the same Task that is in progress.

The downside is that data might have changed in between the two invocations, but the second invocation will not reflect the new changes. Thoughts?

danielgerlag: What if we did not store the result set on this tier at all, the front end could invoke a route that proxies a request for the initial query result through to the view service.

ruokun-niu: Good point. Let me update this

ruokun-niu: Just updated this. The backend now proxies the request to the view service

danielgerlag (approved):

amansinghoriginal (approved):

PR #161, opened 2025-02-28
Transition to Native Multi-Platform Builds for Docker Images

amansinghoriginal:
# Description
This pull request updates our release workflow by moving away from QEMU emulation for multi-platform Docker image builds to using native GitHub runners for linux/amd64 and linux/arm64 architectures. The goal is to enhance build stability and performance while keeping the workflow’s functionality intact.

This change should reduce the build time from 2-3 hours down to < 15 minutes.

### Details
1. Switch to Native Runners
  - Previously, multi-platform builds relied on QEMU emulation. Now, we use:
    - ubuntu-latest runners for linux/amd64.
    - ubuntu-24.04-arm runners for linux/arm64.
  - This removes emulation overhead, leading to fewer build failures and faster execution.

2. Platform-Specific Image Tagging
  - Docker images are now built and tagged individually per architecture:
    - Example: <image>:<tag>-amd64 for linux/amd64.
    - Example: <image>:<tag>-arm64 for linux/arm64.
  - A new manifest creation step combines these into a single multi-platform image (`<image>:<tag>`)
  - Therefore, users should be able to pull images seamlessly without needing to specify the architecture.
 
## Type of change

- [ ] This pull request fixes a bug in Drasi and has an approved issue (issue link required).
- [ ] This pull request adds or changes features of Drasi and has an approved issue (issue link required).
- [x] This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Drasi (issue link optional).

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -19,21 +19,81 @@ on:
     inputs:
       tag:
         description: 'Version Tag'
-        required: true      
+        required: true
       image_prefix:
         description: 'Image Prefix'
         required: false
-        default: 'ghcr.io/drasi-project'  
+        default: 'ghcr.io/drasi-project'
 
 permissions:
-  id-token: write # Required for requesting the JWT
-  contents: read  # Required for actions/checkout
-  packages: read
+  id-token: write
+  contents: read
+  packages: write
 
+# Environment variables with consistent `platforms` fields
 env:
   RELEASE_PATH: ./release
 
+  QUERY_COMPONENTS: '[
+    {"label": "Query Host", "path": "query-container/query-host", "name": "query-container-query-host", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Publish API", "path": "query-container/publish-api", "name": "query-container-publish-api", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "View Service", "path": "query-container/view-svc", "name": "query-container-view-svc", "platforms": "linux/amd64,linux/arm64"}]'
+
+  CONTROL_PLANE_COMPONENTS: '[
+    {"label": "Management API", "path": "control-planes/mgmt_api", "name": "api", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "k8s Resource Provider", "path": "control-planes/kubernetes_provider", "name": "kubernetes-provider", "platforms": "linux/amd64,linux/arm64"}]'
+  
+  SOURCES_COMPONENTS: '[
+    {"label": "Change Router", "path": "sources/shared/change-router", "name": "source-change-router", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Change Dispatcher", "path": "sources/shared/change-dispatcher", "name": "source-change-dispatcher", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Query API", "path": "sources/shared/query-api", "name": "source-query-api", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Debezium Reactivator", "path": "sources/relational/debezium-reactivator", "name": "source-debezium-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "SQL Proxy", "path": "sources/relational/sql-proxy", "name": "source-sql-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "CosmosDB Reactivator", "path": "sources/cosmosdb/cosmosdb-ffcf-reactivator", "name": "source-cosmosdb-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Gremlin Proxy", "path": "sources/cosmosdb/gremlin-proxy", "name": "source-gremlin-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Dataverse Reactivator", "path": "sources/dataverse/dataverse-reactivator", "name": "source-dataverse-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Dataverse Proxy", "path": "sources/dataverse/dataverse-proxy", "name": "source-dataverse-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "EventHub Reactivator", "path": "sources/eventhub/eventhub-reactivator", "name": "source-eventhub-reactivator", "platforms": "linux/amd64,linux/arm64"},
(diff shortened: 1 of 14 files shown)


.github/workflows/draft-release.yml
```diff
@@ -358,21 +375,31 @@ jobs:
...
+            buildx-${{ matrix.component.name }}-${{ matrix.runner }}
 
       - name: Build and Push to GHCR
+        if: contains(matrix.component.platforms, steps.platform.outputs.platform)
         run: |
           cd ${{ matrix.component.path }}
           DOCKER_TAG_VERSION=${{ inputs.tag }} \
           IMAGE_PREFIX=${{ inputs.image_prefix }} \
-          DOCKERX_OPTS="--push --platform ${{ matrix.component.platforms }} --cache-to type=local,dest=/tmp/.buildx-cache,mode=max --cache-from type=local,src=/tmp/.buildx-cache" \
+          DOCKERX_OPTS="--push --platform ${{ steps.platform.outputs.platform }} --cache-to type=local,dest=/tmp/.buildx-cache,mode=max --cache-from type=local,src=/tmp/.buildx-cache" \
```
ruokun-niu, line 402: Correct me if I am understanding this incorrectly, but based on my understanding, doesn't this approach (using matrix) builds the arm and amd images separately? The platform variable is set based on the current runner in the matrix, and the workflow is building one version where `platform` is `linux/amd64` and the other version if `linux/arm64`. 

The reason why I am asking is that I see the latest version of many images are single-arch instead of multi-arch. For example https://github.com/drasi-project/drasi-platform/pkgs/container/api

.github/workflows/draft-release.yml
```diff
@@ -260,30 +270,40 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Determine platform
+        id: platform
+        run: |
+          if [[ "${{ matrix.runner }}" == "ubuntu-latest" ]]; then
+            echo "platform=linux/amd64" >> $GITHUB_OUTPUT
+          elif [[ "${{ matrix.runner }}" == "ubuntu-24.04-arm" ]]; then
```
danielgerlag, line 278: Perhaps we should declare "ubuntu-latest" and "ubuntu-24.04-arm" as constants at the top of the file, so it's easier to change.

.github/workflows/draft-release.yml
```diff
@@ -92,30 +90,41 @@ jobs:
...
+            buildx-${{ matrix.component.name }}-${{ matrix.runner }}
 
       - name: Build and Push to GHCR
         run: |
           cd ${{ matrix.component.path }}
           DOCKER_TAG_VERSION=${{ inputs.tag }} \
           IMAGE_PREFIX=${{ inputs.image_prefix }} \
-          DOCKERX_OPTS="--push --platform linux/amd64,linux/arm64 --cache-to type=local,dest=/tmp/.buildx-cache,mode=max --cache-from type=local,src=/tmp/.buildx-cache" \
-          make 
+          DOCKERX_OPTS="--push --platform ${{ steps.platform.outputs.platform }} --cache-to type=local,dest=/tmp/.buildx-cache,mode=max --cache-from type=local,src=/tmp/.buildx-cache" \
```
danielgerlag, line 116: When the image is pushed, will it not overwrite the existing image tag if it was already written by the other runner?
Eg. 
1. amd64 build starts
2. arm64 build starts
3. amd64 image pushed to ghcr
4. arm64 image pushed to ghcr, overwrites amd64 image

.github/workflows/draft-release.yml
amansinghoriginal: I have rewritten the workflow to generate the single-arch images first and then create the combined manifest.

danielgerlag (approved):

.github/workflows/draft-release.yml
```diff
@@ -92,48 +135,106 @@ jobs:
...
+        component: ${{ fromJson(needs.setup.outputs.control_plane_components) }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # v3.2.0
+        with:
+          submodules: true
+          token: ${{ secrets.DRASI_CORE_PAT }}
```
ruokun-niu, line 237: I think we only need this token when we are trying to build the query container

.github/workflows/draft-release.yml
```diff
@@ -19,21 +19,81 @@ on:
...
+    {"label": "Gremlin Proxy", "path": "sources/cosmosdb/gremlin-proxy", "name": "source-gremlin-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Dataverse Reactivator", "path": "sources/dataverse/dataverse-reactivator", "name": "source-dataverse-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Dataverse Proxy", "path": "sources/dataverse/dataverse-proxy", "name": "source-dataverse-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "EventHub Reactivator", "path": "sources/eventhub/eventhub-reactivator", "name": "source-eventhub-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "EventHub Proxy", "path": "sources/eventhub/eventhub-proxy", "name": "source-eventhub-proxy", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Kubernetes Reactivator", "path": "sources/kubernetes/kubernetes-reactivator", "name": "source-kubernetes-reactivator", "platforms": "linux/amd64,linux/arm64"},
+    {"label": "Kubernetes Proxy", "path": "sources/kubernetes/kubernetes-proxy", "name": "source-kubernetes-proxy", "platforms": "linux/amd64,linux/arm64"}]'
+  
+  REACTIONS_COMPONENTS: '[
```
ruokun-niu, line 61: Sorry I forgot to add the eventbridge reaction when I merged in that PR, can you add it in this component list

ruokun-niu (approved):

.github/workflows/draft-release.yml
amansinghoriginal: The "runs-on" section is not allowing me to reference variables. :/

.github/workflows/draft-release.yml
amansinghoriginal: added

.github/workflows/draft-release.yml
amansinghoriginal: Removed, thanks!

PR #164, opened 2025-03-11
Updated source metadata in the Change Router and Change Dispatcher

ruokun-niu:
# Description

Change Router
- removed the copy of change event in the metadata
- updated start and end timestamp to use nanoseconds.
- Moved the start time to the point when the change router receives the event from pubsub

Change Dispatcher
- updated start and end timestamp to use nanoseconds.
- Moved the start time to the point when the change dispatcher receives the event from pubsub

## Type of change

Fixes: #issue_number

### DIFF
--- a/sources/shared/change-dispatcher/src/main.rs
+++ b/sources/shared/change-dispatcher/src/main.rs
@@ -120,7 +120,7 @@ async fn receive(
     let config = state.config.clone();
     let invoker = &state.invoker;
     let json_data = body["data"].clone();
-    match process_changes(invoker, json_data, config, traceparent,receive_time).await {
+    match process_changes(invoker, json_data, config, traceparent, receive_time).await {
         Ok(_) => StatusCode::OK.into_response(),
         Err(e) => {
             log::error!("Error processing changes: {:?}", e);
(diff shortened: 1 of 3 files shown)


agentofreality (changes requested):
Hi @ruokun-niu, this approach will not work if the Router or Dispatcher receives multiple changes in a single message. Because you set the start time at the very beginning and use it for every output event, they will all have the same start time and appear to take progressively more time for each event.

I like the idea of capturing the start time right at the beginning, but maybe in process_change, you have a local mut start_time that you initialize to the passed value, but then update it for each change.

Doing this, we will capture the overhead of what happens in 'receive' and it will be attributed to the first event if there are multiple, this would make sense for most situations.

ruokun-niu:
> Hi @ruokun-niu, this approach will not work if the Router or Dispatcher receives multiple changes in a single message. Because you set the start time at the very beginning and use it for every output event, they will all have the same start time and appear to take progressively more time for each event.
> 
> I like the idea of capturing the start time right at the beginning, but maybe in process_change, you have a local mut start_time that you initialize to the passed value, but then update it for each change.
> 
> Doing this, we will capture the overhead of what happens in 'receive' and it will be attributed to the first event if there are multiple, this would make sense for most situations.

Just updated both the change router and dispatcher. The first event will use the time when the pubsub initially receives the event, while subsequent events will use the time when each event begins processing.

agentofreality (approved):

amansinghoriginal (approved):

PR #167, opened 2025-03-18
Updated timing for Sources

ruokun-niu:
# Description

This PR updates the timing information for every single available source in Drasi, with changes in both the reactivators and the SDKs. In addition to the source time (when the actual change occurs in the data spource), the reactivator will also now publish the time when it receives the change and the time when it finishes processing the changes. These changes will be parsed and passed by the change router into the subsequent components as metadata. 

The event that is sent by the reactivator and received by the router has a general structure like this:
```json
{
  "op": ...,
  "payload": {
    "after": { ... },
    "before": { ... },
    "source": {
      "db": ...,
      "lsn":  ...,
      "table":  ..., 
      "ts_ns": 1741969672000000000, // source time
    }
  },
  "reactivatorEnd_ns": 8537334252098143252,  // reactivator start time
  "reactivatorStart_ns": 8537334252092701252  // reactivator end time
}
```

The section below goes over how each of these timestamps are deteremined in every Source:

- Relational
  - Source timestamp: This is retrieved from the debezium payload with the `ts_ns` field.
  - Reactivator start time: The timestamp is processed when we receive a new event in the `handleBatch` function, which is a functio in Debezium's `ChangeConsumer` interface
  - Reactivator end time: Right before the processed changes get published
- Cosmos
  - Source timestamp: Retrieved from the feed event, but is only available in seconds
   - Reactivator start time: When the feediterator reads a new event
  - Reactivator end time: Right before the processed changes get published
- Kubernetes
  - Source timestamp: We are using the current system time when the reactivator receives the change. This is the same time as the reactivator start time
  - reactivator start time: described above
  - reactivator end time: Right before the processed changes get published
- EventHub
   - Source timestamp: Uses the enqueued time of the event, however only available in ms
  - reactivator start time: When a partition event is read
  - reactivator end time: Right before the processed changes get published
- Dataverse
  - Source timestamp: Uses the timing from `modifiedon` for `add/update` or thetiming from `deleteTime` for `delete` (100 ns precision)
  - reactivator start time: When a change is read
  - reactivator end time: Right before the processed changes get published

## Type of change

Fixes: #issue_number


query-container/query-host/src/api.rs
```diff
@@ -304,12 +304,12 @@ impl TryInto<SourceChange> for ChangeEvent {
...
                         self.element_type
                             .ok_or(ApiError::BadRequest("missing element type".into()))?,
-                        self.time.ms,
+                        self.time.ns,
                     )?,
             },
             ChangeType::Delete => SourceChange::Delete {
                 metadata: ElementMetadata {
-                    effective_from: self.time.ms,
+                    effective_from: self.time.ns,
```
danielgerlag, line 312: I think the effective from field which is exposed by drasi core is expecting milliseconds. I am not sure how it will behave if you use ns instead?

sources/relational/debezium-reactivator/lib/source.sdk-0.1.4.jar
danielgerlag, line 1: We need a .gitignore entry for this

sources/relational/debezium-reactivator/Dockerfile
```diff
@@ -16,14 +16,52 @@
...
 #
 # Package stage
 #
 FROM --platform=$TARGETPLATFORM mcr.microsoft.com/openjdk/jdk:21-distroless
 COPY --from=build /home/app/target /home/app
-CMD ["-jar","/home/app/debezium-reactivator-1.0.jar"]
\ No newline at end of file
+CMD ["-jar", "/home/app/debezium-reactivator-1.0.jar"]
+
+# Copyright 2024 The Drasi Authors.
```
danielgerlag, line 39: Did you mean to remove this section?

sources/sdk/dotnet/Drasi.Source.SDK/Models/SourceChange.cs
```diff
@@ -33,19 +35,22 @@ public class SourceChange
...
+        this.tsNS = timestampNs;
+        this.reactivatorStartNs = reactivatorStartNs;
         this.element = element;
         this.lsn = lsn;
         this.partition = partition;
     }
 
     public string ToJson()
     {
+        var reactivatorEndNs = (DateTimeOffset.UtcNow.Ticks - DateTimeOffset.UnixEpoch.Ticks) * 100;
```
danielgerlag, line 53: I'm not sure we can assume that when someone calls .ToJson() as the end of the reactivator leg.

sources/sdk/java/src/main/java/io/drasi/source/sdk/models/SourceChange.java
```diff
@@ -93,7 +93,9 @@ public String toJson() {
                 result.put("op", "d");
                 break;
         }
-        result.put("ts_ms", tsMS);
+        result.put("reactivatorStart_ns", tsNS);
+        long currentTime = System.nanoTime();
+        result.put("reactivatorEnd_ns", currentTime);
```
danielgerlag, line 98: I'm not sure we should assume the serialization time is the same as reactivatorEnd_ns

sources/relational/debezium-reactivator/Dockerfile
ruokun-niu: Yeah I will revert the changes made to the dockerfile once the new version of the SDK is published

query-container/query-host/src/api.rs
ruokun-niu: Right. Should we convert this back to the ms, and keep the timings in metadata to use ns?

danielgerlag: We need to test this with drasi core. I think internally it is a u64, not sure if nanoseconds can always fit into that size field?

sources/sdk/dotnet/Drasi.Source.SDK/Models/SourceChange.cs
ruokun-niu: How about we move the end time to here https://github.com/drasi-project/drasi-platform/blob/main/sources/sdk/dotnet/Drasi.Source.SDK/Reactivator.cs#L60-L63, before the SourceChange gets published?

query-container/query-host/src/api.rs
ruokun-niu: I'll look into this and update my findings here

sources/sdk/dotnet/Drasi.Source.SDK/Models/SourceChange.cs
ruokun-niu: I have moved this out of ToJson(). The endtime is determined before we call `publisher.publish(change)`. See https://github.com/ruokun-niu/drasi-platform/blob/reactivator-timing/sources/sdk/dotnet/Drasi.Source.SDK/Reactivator.cs#L64

sources/sdk/java/src/main/java/io/drasi/source/sdk/models/SourceChange.java
ruokun-niu: Will update this to follow the dotnet SDK, where the end time is determined before `publish`

ruokun-niu: I think there are two approaches to this
1. In the `Publish` function of the change publisher, we determine the end time and update the source change (https://github.com/drasi-project/drasi-platform/blob/d343738f4412c0a099ea047b5577448d59f709f4/sources/sdk/java/src/main/java/io/drasi/source/sdk/DaprChangePublisher.java#L41). This allows us to keep the logic for determining the end time within the SDK, and the users won't have to worry about it. This means that the endtime will only be determined when we publish the change event. As of right now, I can't think of any other situation where we need to determine an end timestamp but do not publish the change.
2. Leave it to the users to determine when to specify the end time. They are responsible for doing this when implementing `run()`

Personally I am leaning towards approach #1, any thoughts @danielgerlag

query-container/query-host/src/api.rs
ruokun-niu: One issue with using nanoseconds is that in drasi-core, all of our timestamps are measured in milliseconds (and uses `u64`). Similarly, time-related functions such as `getVersionByTimestamp` also work with milliseconds instead of nanoseconds. I believe a significant amount of refactoring would be required if we decide to convert effective_from to use nanoseconds.

ruokun-niu: A simpler workaround would be to keep the change event time with millisecond-precision in `time` field of the ChangeEvent (this timing is used to construct the effective_from). and store the source timestamp with nanosecond-precision in the metadata. 

For query processing in drasi-core, we will continue to work with milliseconds. The source timestamp is still available in the metadata for other purposes such as tracing

danielgerlag: sounds good 👍

ruokun-niu: Updated this field to use ms and reverted the changes madein the query host

sources/sdk/java/src/main/java/io/drasi/source/sdk/models/SourceChange.java
ruokun-niu: updated

agentofreality: @ruokun-niu I also agree that option 1 was the best way.

sources/dataverse/dataverse-reactivator/Services/JsonEventMapper.cs
```diff
@@ -85,16 +86,16 @@ public Task<ChangeNotification> MapEventAsync(IChangedItem rawEvent)
                         var dt = delete.RemovedItem.KeyAttributes["deletetime"];
                         if (dt is DateTime time)
                         {
-                            result.TimestampMilliseconds = (long)(time.ToUniversalTime() - new DateTime(1970, 1, 1)).TotalMilliseconds;
+                            result.TimestampNanoseconds = (long)(time.ToUniversalTime() - new DateTime(1970, 1, 1)).TotalMilliseconds;
                         }
                         else if (dt is DateTimeOffset offset)
                         {
-                            result.TimestampMilliseconds = offset.ToUnixTimeMilliseconds();
+                            result.TimestampNanoseconds = (offset.UtcTicks - DateTimeOffset.UnixEpoch.Ticks) / 10000;
```
agentofreality, line 93: Is this correct? Why divide by 10000?

sources/shared/change-router/src/main.rs
```diff
@@ -540,18 +539,20 @@ async fn process_changes(
                     "subscriptions": subscriptions,
                     "time": {
                         "seq": change["payload"]["source"]["lsn"],
-                        "ms": change["ts_ms"]
+                        "ms": change["payload"]["source"]["ts_ns"].as_u64().unwrap_or(0) / 1_000_000,  // convert to milliseconds for drasi-core
```
agentofreality, line 542: We cant just default to zero here in case of conversion issues. If we expect it will never happen, then we should log error and not pass the record on. A zero time will cause many downstream problems, and might be hard to find if it is getting introduced here.

sources/dataverse/dataverse-reactivator/Services/JsonEventMapper.cs
```diff
@@ -85,16 +86,16 @@ public Task<ChangeNotification> MapEventAsync(IChangedItem rawEvent)
                         var dt = delete.RemovedItem.KeyAttributes["deletetime"];
                         if (dt is DateTime time)
                         {
-                            result.TimestampMilliseconds = (long)(time.ToUniversalTime() - new DateTime(1970, 1, 1)).TotalMilliseconds;
+                            result.TimestampNanoseconds = (long)(time.ToUniversalTime() - new DateTime(1970, 1, 1)).TotalMilliseconds;
```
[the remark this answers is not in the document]
ruokun-niu: good catch, thanks

sources/dataverse/dataverse-reactivator/Models/ChangeNotification.cs
```diff
@@ -93,8 +97,8 @@ public class ChangeSource
         [JsonPropertyName("lsn")]
         public long LSN { get; set; }
 
-        [JsonPropertyName("ts_ms")]
-        public long TimestampMilliseconds { get; set; }
+        [JsonPropertyName("ts_ns")]
```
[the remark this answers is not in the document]
ruokun-niu: I think we should remove the ts_sec field. This field was never used in the original implementation and I don't think there is a need to keep a ns_timestamp, a ms_timestamp and a sec_timestamp as separate fields

sources/dataverse/dataverse-reactivator/Services/JsonEventMapper.cs
ruokun-niu: fixed

query-container/query-host/src/change_stream/redis_change_stream.rs
```diff
@@ -308,6 +308,25 @@ where
...
+                        }
+                    },
+                    Err(err) => {
+                        log::error!("Failed to deserialize enqueue_time to String: {:?}", err);
+                        0
+                    }
+                },
+                _ => 0,
+            },
+            None => 0,
```
danielgerlag, line 328: Should it be null if we don't know the enqueue time?

danielgerlag (commented):
Do we also need to update anything in the typespec definitions?

ruokun-niu:
> Do we also need to update anything in the typespec definitions?

Good point. I have updated the typespec definitions for `source-output`. For `query_output`, we do not need to update the typespec because the timing information is stored in the metadata, which is already defined

agentofreality (approved):

query-container/publish-api/src/publisher.rs
```diff
@@ -61,7 +61,7 @@ impl Publisher {
         let mut connection = self.connection.clone();
         let mut items = Vec::with_capacity(4);
         let now = match SystemTime::now().duration_since(SystemTime::UNIX_EPOCH) {
-            Ok(now) => now.as_millis().to_string(),
+            Ok(now) => now.as_nanos().to_string(),
```
danielgerlag, line 64: are we storing this value as a number or a string?

query-container/query-host/src/change_stream/redis_change_stream.rs
```diff
@@ -308,6 +308,34 @@ where
...
+                    },
+                    Err(err) => {
+                        return Err(ChangeStreamError::MessageError {
+                            id: message.id.clone(),
+                            error: format!("Failed to deserialize enqueue_time: {:?}", err),
+                        });
+                    }
+                },
+                _ => {
+                    return Err(ChangeStreamError::MessageError {
```
danielgerlag, line 331: Not sure if this should be a warning or error.

query-container/publish-api/src/publisher.rs
ruokun-niu: Good catch. I think it was stored as a string in the previous implementation, but we should be able to store it as a number instead. Updating this

ruokun-niu: Oh I think it is set as a string because `items` is expecting a `String`

PR #180, opened 2025-04-01
Fixed Rust Source SDK tests and updated the workflow & makefiles for testing Source SDKs

ruokun-niu:
# Description

- Fixed the tests for the Rust Source SDK
- Updated the makefile for each Source SDK to run tests before publishing
- Updated the GH workflow to test the Source SDKs

## Type of change

Fixes: #issue_number


sources/sdk/dotnet/Makefile
```diff
@@ -1,11 +1,11 @@
 .PHONY: install-dependencies test package
 
 install-dependencies:
-	dotnet restore
+	cd Drasi.Source.SDK && dotnet restore
```
danielgerlag, line 4: the dotnet toolchain should work with the .sln file in this directory?

typespec/source-output/main.tsp
```diff
@@ -17,8 +17,8 @@ enum Versions {
 @discriminator("op")
 model SourceChange {
     "op": "i" | "u" | "d";
-    reactivatorStart_ns: int64;
-    reactivatorEnd_ns: int64;
+    reactivatorStart_ns: numeric;
+    reactivatorEnd_ns: numeric;
```
danielgerlag, line 21: How does this effect the code generation?

ruokun-niu: Good point. I looked into types for typespec a bit more and I believe that using a super type like `numeric` might not be appropriate. I think we should use uint64 instead of int64, as 1. timestamp will never be negative and 2. uint64 has a higher range than int64

ruokun-niu: For C#: `int64` generated `long`, `numeric` gave `decimal`, but `uint64` generates `ulong`. This keeps it an integer type (unlike `decimal`), just unsigned.
- **Rust**: `int64` → `i64`, `uint64` → `u64`

ruokun-niu: An alternative is to use uint64 for the typespec, and change the respective types in the SDKs (e.g. C#: long -> ulong)

danielgerlag (approved):

PR #189, opened 2025-04-09
Observability stack for drasi-platform

ruokun-niu:
# Description

Added a new flag called `observability-level` to the `drasi init` command. There are four possible values:
- `none`: Default flag. No observability infra will be deployed
- `tracing`: Installs tempo, grafana and otel
-  `metrics`: installs prometheus, grafana and otel
- `full`:  Installs everything

## Type of change

Fixes: #issue_number


cli/service/resources/observability/full-observability.yaml
```diff
@@ -0,0 +1,256 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: grafana
+spec:
+  ports:
+  - port: 3000
+    targetPort: 3000
+    name: http
+  type: NodePort
```
amansinghoriginal, line 254: Why do we use NodePort for Grafana? I guess it will work for local kind/k3d.
Should we consider Ingress, or is that overkill for now.

cli/service/resources/observability/metrics.yaml
```diff
@@ -0,0 +1,151 @@
...
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/prometheus
+        - name: data-volume
+          mountPath: /prometheus
+      volumes:
+      - name: config-volume
+        configMap:
+          name: prometheus-config
+      - name: data-volume
```
amansinghoriginal, line 57: Doesn't using `emptyDir` mean that the data will be lost on pod-restart? This is for both prometheus & tempo.
Should we implement PersistentVolumeClaims similar to how Redis and MongoDB are configured?

amansinghoriginal (commented):
Looks good overall, had a couple of questions.

cli/cmd/init.go
```diff
@@ -96,7 +97,33 @@ Usage examples:
 			if err != nil {
 				return err
 			}
-			if err := installer.Install(local, registry, version, output, namespace, daprRegistry); err != nil {
+			observabilityLevel, err := cmd.Flags().GetString("observability-level")
+			if err != nil {
+				return err
+			}
+			// Validate observability-level
```
amansinghoriginal, line 104: Thanks for validating the options, and making the options structured.

cli/service/resources/observability/metrics.yaml
ruokun-niu: Good point. I didn't add support for this initially as I just wanted to get the tools working. Will look into this

cli/service/resources/observability/full-observability.yaml
ruokun-niu: Right now I think we can just rely on port-forwarding as a starting point

cli/service/resources/observability/metrics.yaml
ruokun-niu: I have updated the deployments to use a PVC.

cli/service/resources/observability/full-observability.yaml
amansinghoriginal: Alright

amansinghoriginal (approved):
LGTM

danielgerlag (approved):

PR #240, opened 2025-05-26
Use Renovate Github App instead of Dependabot

amansinghoriginal:
# Description

Dependabot Config is not working with wildcards. Instead of dependabot, we can use [Renovate Github App](https://github.com/apps/renovate). I have requested for the bot which once approved by Drasi Owners, should enable the renovate bot to run at the set schedule.

This should reduce the number of pull requests created as patches and minor bumps are grouped together.

- We will consider only patch updates that are older than 7 days, and all of them will be grouped together into a single PR. This will remain open for 3 days, and then automatically gets merged by a github action.
- We will consider only minor updates taht are older than 14 days, and all of them will be grouped together into a single PR. This will remain open for 10 days before getting automatically merged by the github action.
- Each major version update will get its own PR.
- At any one time, renovate will open a max of 10 PRs.

- We can prevent automatic merging of a PR by simply removing the `candidate label` added by renovate, or we can close the PR.

## Type of change
- This pull request is a maintenance task and doesn't change the functionality of Drasi.


danielgerlag (approved):

ruokun-niu (approved):

.github/renovate.json
```diff
@@ -0,0 +1,39 @@
...
+  "vulnerabilityAlerts": {
+    "enabled": true,
+    "labels": ["security", "vulnerability"]
+  },
+  "packageRules": [
+    {
+      "description": "Group stable patch updates for potential automerge",
+      "matchUpdateTypes": ["patch"],
+      "minimumReleaseAge": "7 days",
+      "automerge": true,
```
ruokun-niu, line 19: Quick question, based on my understanding, this setting should automerge patch updates for a specific package together into one fix. In this case, what does the `automerge.yml` file in this PR do? Is it for merging multiple PRs for the same package/component into one PR?

amansinghoriginal: You're right, let's turn this off.

Renovate supports a schedule for auto-merging, but it doesn't seem to support waiting for X days before merging the open PR.

ruokun-niu: Sounds good. Maybe this could be something that we can add later based on our experience of using renovate

PR #253, opened 2025-06-10
Initial http reaction

danielgerlag:
# Description

Initial implementation of an Http reaction. This enables the user to craft Http calls that will be invoked when query result set changes occur.

The URL of the request and the payload can be constructed from a [handlebars template](https://handlebarsjs.com/).

For example, the following configuration will `post` to `https://api.github.com/repos/{{repo}}/issues/{{issue_number}}/comments`, where the variables are sourced from the result set item that was added. The JSON body of the request is also constructed using handlebars.

```yaml
kind: Reaction
apiVersion: v1
name: test-http
spec:
  kind: Http
  properties:
    baseUrl: "https://api.github.com"
    token: xxxx
  queries:
    issue-comments: >
      addedResult:
        url: "/repos/{{repo}}/issues/{{issue_number}}/comments"
        method: "POST"
        body: > 
          {
            "body": "Hello! @{{creator}}"
          }
        headers:
          X-GitHub-Api-Version: "2022-11-28"
      updatedResult: 
            ....
      deletedResult: 
            ....
```

# What is missing
- Currently, we do not have support for resilience/retry strategies.
- Currently, we only support bearer token auth for the target endpoint.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -311,6 +311,11 @@ jobs:
             label: 'PostDaprPubSub',
             path: './reactions/dapr/post-pubsub', 
             name: 'reaction-post-dapr-pubsub'
+          },
+          { 
+            label: 'Http',
+            path: './reactions/http', 
+            name: 'reaction-http'
           }
         ]
     steps:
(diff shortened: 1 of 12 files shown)


amansinghoriginal:
> > Can you create an issue for this with various config options supported?
> 
> Added to the PR description.

Yes, thanks for adding it. Can we please also create issue - as anyways for major changes we require the PR to link to an issue.
An added benefit is that OSS community can easily track what's going on in the project - and we can also create a board from issues - so better tracking?

PR #256, opened 2025-06-13
Updated the start id of the Redis Change Stream in the Query Host

ruokun-niu:
# Description

Currently, the Redis Change Stream in the Query Host uses `"$"` as the start id for the redis stream. In some situations where changes begin flowing in before the bootstrap completes, some early entries will be missed. This PR updates the Change stream to accept an optional start id as one of the inputs. When the query worker starts, it will record a timestamp at the start (before bootstrap), and use that as the start_id for the redis change stream

## Type of change

Fixes: #issue_number


query-container/query-host/src/change_stream/redis_change_stream.rs
```diff
@@ -40,12 +40,24 @@ impl RedisChangeStream {
...
+            },
+            None => {
+                log::info!("No start_id provided, using '$'");
+                "$"
+            }
+        };
+        
         match connection
-            .xgroup_create_mkstream::<&str, &str, &str, String>(topic, group_id, "$")
+            .xgroup_create_mkstream::<&str, &str, &str, String>(topic, group_id, starting_position)
```
danielgerlag: I'm not sure this redis command works with a timestamp, I think it's an ID?
https://redis.io/docs/latest/commands/xgroup-create/

query-container/query-host/src/query_worker.rs
```diff
@@ -146,6 +146,13 @@ impl QueryWorker {
                 }
             };
 
+            let start_timestamp = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_millis();
+            
+            let start_id = format!("{}-0", start_timestamp);
```
danielgerlag: This is a Redis specific formatted ID. We should hide the Redis specific bits in the redis stream implementation.

query-container/query-host/src/change_stream/redis_change_stream.rs
ruokun-niu: Yes, starting_position  is consturcted with `{tiemstamp}-0`, which is a properly formatted Redis Stream ID.

query-container/query-host/src/query_worker.rs
ruokun-niu: Good point. Perhaps we can modify the redis change stream to accept a timestamp string, and then create a Redis ID in `new()`

danielgerlag: Would it be a string or a number?

ruokun-niu: I think as a number? The redis stream will format it into an ID

query-container/query-host/src/query_worker.rs
```diff
@@ -146,6 +146,11 @@ impl QueryWorker {
                 }
             };
 
+            let start_timestamp = std::time::SystemTime::now()
```
danielgerlag, line 149: What would the behaviour be for queries that are restarting, or moving from one query host to another, rather than starting for the first time?

ruokun-niu: I think when the query restarts or when the query moves to a different query host replica, we will continue to use the previously created consumer group and use that group's stored position. (https://github.com/ruokun-niu/drasi-platform/blob/query-worker-fix/query-container/query-host/src/change_stream/redis_change_stream.rs#L63-L72)

However, let's say that the query gets deployed to a different query container, a new consumer group will be created, since the Redis topic is formatted from the query container ID

danielgerlag: Could we verify this behaviour?

ruokun-niu: I'll test this

ruokun-niu: Just tested this. When the query restarts, it resumes from the stored consumer group position. However when we mobve to a new query container, a new consumer group was created

danielgerlag: I think that is fine because moving to another query container would mean reading a different stream altogether.
I meant, moving between hosts within the same query container.

ruokun-niu: Yep, makes sense

query-container/query-host/src/change_stream/redis_change_stream.rs
```diff
@@ -40,12 +40,18 @@ impl RedisChangeStream {
         group_id: &str,
         buffer_size: usize,
         fetch_batch_size: usize,
+        start_timestamp: Option<u128>,
```
agentofreality: Is there any reason to make this Option<u128>, do we ever need to use "$" ?

ruokun-niu: In the query host, the RedisChangeStream is only used in the Query Worker. I can modify `start_timestamp` to accept u128 instead of Option<u128>.

danielgerlag (approved):

amansinghoriginal (approved):

agentofreality (approved):

PR #286, opened 2025-08-10
Add Query Language Support for GQL and Cypher

ama2369:
## Summary

This PR adds support for specifying query language in continuous query YAML configuration. Users can now select between Cypher and GQL languages using the optional `queryLanguage` property.

## Usage

Users can now specify query language in their continuous query YAML:

```yaml
apiVersion: v1
kind: ContinuousQuery
spec:
  mode: query
  queryLanguage: Cypher
  query: "MATCH (n) RETURN n"
  # ... other configuration
```

Or for GQL:

```yaml
apiVersion: v1
kind: ContinuousQuery
spec:
  mode: query
  queryLanguage: GQL
  query: "MATCH (n) RETURN n"
  # ... other configuration
```

Existing queries without `queryLanguage` property continue to default to Cypher


control-planes/mgmt_api/src/api/v1/models/query.rs
```diff
@@ -51,6 +51,7 @@ pub struct QuerySpecDto {
     pub container: String,
     pub mode: String,
     pub query: String,
+    pub query_language: Option<String>,
```
danielgerlag: Let's make an enum for query language

ruokun-niu:
+1 to what Daniel said. Otherwise the PR looks good.

Also I noticed that the build workflow for query host is failing. My guess is that the workflow is trying to pull the `main` branch of `drasi-core` as a submodule. You might want to update it to point to your feature branch temporarily to pass the build test. We can update it later when we are merging things into `main`

control-planes/mgmt_api/src/api/v1/models/query.rs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use resource_provider_api::models::QueryLanguage;
```
danielgerlag: The `API` layer is something that is exposed externally and we would want to maintain backwards compatibility and/or multiple versions in the future. For that reason, the internals of the service need to be able to evolve independently from the API, so we should avoid importing artifacts from other internal layers in the external API. We should declare the enum again, specifically for the API layer and then have a mapping between it and the domain layer. This is all to keep the external contract (API) clean from an ever evolving internal implementation.

PR #287, opened 2025-08-11
Add ingress support for Drasi sources and reactions

ruokun-niu:
# Description

This pull request introduces ingress support for Drasi Sources and Reactions, enabling external access through Contour ingress controller. The changes include a new CLI command for managing ingress, integration of Contour installation and configuration, updates to RBAC permissions, and enhancements in both the CLI and Kubernetes provider to handle ingress resources and expose endpoints externally.

To use the ingress functionality, user would first have to install the Contour Ingress controller by using the new command `drasi ingress init`.  This command uses helm chart to install the necessary resources in the `projectcontour` namespace.

When a user creates a Source/Reaction that has a service with an endpoint where the setting is set to `external`, a Kubernetes service and ingress will both be created. The URL will be output in the terminal after `drasi apply` completes.

Example:
```bash
drasi apply -f reaction.yaml
✓ Apply: Reaction/hello-world-debug: complete
Ingress URL: http://hello-world-debug.drasi.4.246.5.128.nip.io
```

Below is a detailed list of all of the changes I've made:

**Ingress support and CLI enhancements:**

* Added `NewIngressCommand` to the CLI, allowing users to initialize and manage Contour ingress for Drasi Sources and Reactions. This includes automated installation of Contour via Helm and configuration for external access. (`cli/cmd/ingress.go`, `cli/cmd/root.go`) [[1]](diffhunk://#diff-1a86daabac6fe35c61cea24e6eae7f92b70d4a961fd0744a26601e02d9cad707R1-R101) [[2]](diffhunk://#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5R38)
* Implemented `ContourInstaller` in the SDK, handling Helm-based installation and setup of Contour in the `projectcontour` namespace. (`cli/sdk/ingress/contour_installer.go`)

**RBAC and resource configuration:**

* Updated RBAC roles and bindings to grant necessary permissions for managing ingress resources and accessing the Contour service. This includes new rules for `ingresses` and a `ClusterRole` for the `contour-envoy` service. (`cli/installers/resources/service-account.yaml`) [[1]](diffhunk://#diff-e865e8f1b24bdd3cbb361481a4307c66d4cd7bb51245093e62006baeaa46f747R109-R119) [[2]](diffhunk://#diff-e865e8f1b24bdd3cbb361481a4307c66d4cd7bb51245093e62006baeaa46f747R132-R159)
* Modified reaction provider resource definitions to use `service` instead of `gateway` for endpoint configuration, aligning with the new ingress model. (`cli/installers/resources/default-reaction-providers.yaml`)

**Kubernetes provider integration:**

* Extended the Rust Kubernetes provider to manage ingress resources: added support for creating, updating, and deleting ingresses, and dynamically generates external hostnames using the Contour LoadBalancer IP. (`control-planes/kubernetes_provider/src/controller/reconciler.rs`) [[1]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R26) [[2]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R51) [[3]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R88) [[4]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R243-R260) [[5]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R445-R526) [[6]](diffhunk://#diff-0c617fe6c468eef9203813c4ccbea34c72120a83a92ffd780310821c4ec977b1R620-R623)

**User experience improvements:**

* Enhanced CLI apply workflow to display ingress URLs for reactions and sources with external endpoints, automatically discovering the Contour LoadBalancer IP. (`cli/sdk/api_client.go`) [[1]](diffhunk://#diff-e761e1cd17c8c0b5f051f1ae046e2c8f6cb3144079f70f0167c4375275e13e12R19) [[2]](diffhunk://#diff-e761e1cd17c8c0b5f051f1ae046e2c8f6cb3144079f70f0167c4375275e13e12R31-R33) [[3]](diffhunk://#diff-e761e1cd17c8c0b5f051f1ae046e2c8f6cb3144079f70f0167c4375275e13e12R92-R95) [[4]](diffhunk://#diff-e761e1cd17c8c0b5f051f1ae046e2c8f6cb3144079f70f0167c4375275e13e12R262-R346)

These changes collectively enable seamless ingress management and external access for Drasi workloads in Kubernetes environments.

## Type of change

Fixes: #issue_number

### DIFF
--- a/cli/AGENTS.md
+++ b/cli/AGENTS.md
@@ -1,0 +1,97 @@
+# AGENTS.md: Drasi CLI
+
+This document provides a high-level overview of the Drasi Command Line Interface (CLI) project for coding agents. The CLI is a Go application built with the Cobra library, responsible for installing and managing Drasi environments and resources.
+
+The `README.md` file contains additional context intended for human developers. For a detailed command reference, use the CLI's built-in `help` command (e.g., `drasi help apply`).
+
+## Core Workflow
+
+A typical command execution follows this flow:
+1.  The user runs a command defined in the **`cmd/`** directory.
+2.  The command uses the **`sdk/`** to get a `PlatformClient` for the target environment.
+3.  The `PlatformClient` establishes a connection (e.g., a Kubernetes port-forward) and provides an `ApiClient`.
+4.  The command uses the `ApiClient` to make HTTP calls to the Drasi Management API.
+5.  Throughout the process, the **`output/`** package is used to render status and results to the user's terminal.
+
+## Component Diagram
+
+This diagram shows the structural dependencies between the major packages in the CLI.
+
+```mermaid
+graph TD
+    subgraph "Entry Point"
+        main("main.go")
+    end
+
+    subgraph "Command Layer"
+        Cmd("cmd/")
+    end
+
+    subgraph "Service & Abstraction Layers"
+        SDK("sdk/")
+        Installers("installers/")
+    end
+    
+    subgraph "Utility & Data Packages"
+        Output("output/")
+        API("api/")
+        Config("config/")
+        Registry("sdk/registry/")
+    end
+
+    main --> Cmd
+    
+    Cmd --> SDK
+    Cmd --> Installers
+    Cmd --> Output
+    Cmd --> Config
+
+    Installers --> SDK
+    Installers --> API
+
+    SDK --> API
+    SDK --> Registry
+```
+
+## Project Structure
+
+The CLI's functionality is organized into the following key directories. Each contains a local `AGENTS.md` file for more detailed information.
+
+-   **`api/`**: Defines the core Go structs (`Manifest`, `Resource`) that represent Drasi resources in YAML files and in API communications. This is the foundational data model for the CLI.
+-   **`cmd/`**: Contains the implementation for every user-facing command (e.g., `drasi apply`, `drasi init`). This is the entry point for all CLI functionality and orchestrates calls to the `sdk` and `installers`.
+-   **`config/`**: Defines and initializes global, build-time variables like `Version` and `Registry`.
+-   **`installers/`**: Holds the logic for installing and uninstalling Drasi environments. It abstracts platform specifics (Kubernetes vs. Docker) and contains the embedded YAML manifests for deployment in its `resources/` subdirectory.
+-   **`output/`**: Manages all terminal output. It provides a rich, interactive UI for terminals and falls back to plain-text logging for non-interactive environments.
+-   **`sdk/`**: The core abstraction layer. It provides a `PlatformClient` to handle environment-specific operations (like port-forwarding in Kubernetes) and an `ApiClient` to communicate with the Drasi Management API.
+
+## Key Abstractions
+
+-   **`sdk.PlatformClient`**: An interface that decouples the CLI commands from the underlying platform (e.g., Kubernetes). Implementations handle tasks like creating API clients, managing tunnels, and interacting with the platform's secret store. See `sdk/platform_client.go`.
+-   **`installers.Installer`**: An interface for the multi-step process of deploying a Drasi environment. The `KubernetesInstaller` is the primary implementation. See `installers/installer.go`.
+-   **`output.TaskOutput`**: An interface for rendering progress to the user, with implementations for rich terminal UIs and simple log output. See `output/interface.go`.
+
+## Build and Test
+
+The project is built using Go and a `Makefile`.
+
+-   **Build for all platforms**:
+    ```bash
+    make
+    ```
+-   **Build for the current platform**:
+    ```bash
+    make build
+    ```
+-   **Install a local build**:
+    ```bash
+    # macOS / Linux
+    sudo make install
+
+    # Windows (in an elevated terminal)
+    make install
+    ```
+-   **Formatting and Linting**:
+    ```bash
+    make fmt
+    make vet
+    ```
--- a/cli/CLAUDE.md
+++ b/cli/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/cli/api/AGENTS.md
+++ b/cli/api/AGENTS.md
@@ -1,0 +1,18 @@
+# AGENTS.md: `cli/api` Directory
+
+This directory contains the core Go data structures that represent Drasi resources and manifests. These types are used by the CLI to parse resource definition files and to serialize/deserialize data when communicating with the Drasi Management API.
+
+## Files
+
+### `manifest.go`
+
+-   **Purpose**: Defines the structure for a Drasi resource manifest, which is how resources are defined in YAML files. It provides a utility function to read and parse one or more YAML documents from a byte slice.
+-   **Key Components**:
+    -   `Manifest` (struct): Represents a single resource definition from a YAML file. It includes fields like `Kind`, `ApiVersion`, `Name`, and `Spec`. This is the primary structure used when parsing files for `drasi apply` or `drasi delete`.
+    -   `ReadManifests([]byte)` (function): A parser that takes raw byte data (from a YAML file) and decodes it into a slice of `Manifest` structs, correctly handling multi-document YAML files.
+
+### `resource.go`
+
+-   **Purpose**: Defines the generic structure for a Drasi resource as it is represented by the Drasi Management API.
+-   **Key Components**:
+    -   `Resource` (struct): A generic representation of any Drasi resource (Source, Reaction, Query, etc.). It contains the `Id`, the configuration (`Spec`), and the current operational state (`Status`) of the resource. This struct is used for handling responses from the management API, such as for the `drasi describe` command.
--- a/cli/api/CLAUDE.md
+++ b/cli/api/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/cli/cmd/AGENTS.md
+++ b/cli/cmd/AGENTS.md
@@ -1,0 +1,39 @@
+# AGENTS.md: `cli/cmd` Directory
+
+This directory contains the implementation for each command available in the Drasi CLI (e.g., `apply`, `list`, `init`). It uses the `Cobra` library to structure the commands, arguments, and flags. Each file typically defines a single top-level command and its associated logic.
+
+The general pattern for each command is:
+1.  Define the `cobra.Command` struct with its `Use`, `Short`, and `Long` descriptions.
+2.  Implement the `RunE` function, which contains the core logic for the command.
+3.  Inside `RunE`, parse flags, load the current environment configuration (`registry.LoadCurrentRegistration`), create a platform client (`sdk.NewPlatformClient`), and execute the desired action.
+
+## Files
+
+### `root.go`
+
+-   **Purpose**: This is the entry point for the CLI. It creates the root `drasi` command and attaches all other commands (from the other files in this directory) to it.
+-   **Key Components**:
+    -   `MakeRootCommand()`: Initializes the main `drasi` command and registers all subcommands. It also defines persistent flags like `--namespace` that are available globally.
+
+### Resource Management Commands
+
+-   **`apply.go`**: Implements the `drasi apply` command. It loads resource manifests from files (`-f` flag) and sends them to the Drasi Management API to be created or updated.
+-   **`delete.go`**: Implements the `drasi delete` command. It can delete resources specified by `[kind] [name]` arguments or from manifest files (`-f` flag).
+-   **`describe.go`**: Implements the `drasi describe` command. It fetches and displays the detailed configuration (`spec`) and current `status` of a single, specified resource.
+-   **`list.go`**: Implements the `drasi list` command. It retrieves a list of resources of a given kind and displays their status in a table format.
+-   **`wait.go`**: Implements the `drasi wait` command. It blocks until specified resources become fully operational or a timeout is reached. This is useful for scripting.
+
+### Installation & Environment Commands
+
+-   **`init.go`**: Implements the `drasi init` command. This is a complex command responsible for installing Drasi into a Kubernetes cluster or a local Docker container. It handles various flags for customization (`--local`, `--version`, `--registry`, `--manifest`, etc.).
+-   **`uninstall.go`**: Implements the `drasi uninstall` command. It removes a Drasi installation by deleting its Kubernetes namespace and can optionally remove Dapr as well.
+-   **`env.go`**: Implements the `drasi env` command and its subcommands (`all`, `current`, `delete`, `kube`, `use`). It manages saved connection configurations for different Drasi environments.
+-   **`namespace.go`**: Implements the `drasi namespace` command and its subcommands (`get`, `set`, `list`). It manages the default Kubernetes namespace that the CLI targets.
+
+### Utility & Debugging Commands
+
+-   **`watch.go`**: Implements the `drasi watch` command. It connects to a query's real-time view service and displays a continuously updating table of its result set.
+-   **`tunnel.go`**: Implements the `drasi tunnel` command. It creates a local port forward to a service running inside the Kubernetes cluster, enabling local debugging.
+-   **`secret.go`**: Implements the `drasi secret` command and its subcommands (`set`, `delete`). It provides a way to manage Kubernetes secrets for storing sensitive data.
+-   **`version.go`**: Implements the `drasi version` command. It's a simple command that prints the compiled-in version of the CLI.
+-   **`utils.go`**: Contains helper functions used by other commands in this directory, notably `loadManifests`, which is responsible for reading resource definitions from files, URLs, or standard input.
--- a/cli/cmd/CLAUDE.md
+++ b/cli/cmd/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/cli/config/AGENTS.md
+++ b/cli/config/AGENTS.md
@@ -1,0 +1,12 @@
+# AGENTS.md: `cli/config` Directory
+
+This directory contains global configuration for the Drasi CLI application.
+
+## Files
+
+### `config.go`
+
+-   **Purpose**: Defines and initializes global, package-level variables for the CLI. These variables are typically set at build time to inject versioning and other important information into the binary.
+-   **Key Components**:
+    -   `Version` (string): Stores the version of the Drasi CLI. It defaults to `"latest"` if not set during the build process. This is the version displayed by the `drasi version` command and used by default in `drasi init`.
+    -   `Registry` (string): Stores the default container registry path to pull Drasi images from. It defaults to `"ghcr.io"` if not set during the build. This is used by the `drasi init` command.
--- a/cli/config/CLAUDE.md
+++ b/cli/config/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/cli/installers/AGENTS.md
+++ b/cli/installers/AGENTS.md
@@ -1,0 +1,65 @@
+# AGENTS.md: `cli/installers` Directory
+
+This directory contains the logic for installing, uninstalling, and managing Drasi environments on different platforms (primarily Kubernetes). It abstracts the platform-specific details of deployment behind a set of interfaces. The `drasi init` and `drasi uninstall` commands delegate their core work to the components in this directory.
+
+## Core Interfaces
+
+### `installer.go`
+
+-   **Purpose**: Defines the main `Installer` interface. This provides a generic contract for installing Drasi, regardless of the target platform.
+-   **Key Components**:
+    -   `Installer` (interface): The core interface with the `Install(...)` method.
+    -   `MakeInstaller(...)` (factory function): A factory function that takes a `registry.Registration` (which defines the target environment) and returns the appropriate concrete `Installer` implementation (e.g., `KubernetesInstaller`).
+
+### `uninstaller.go`
+
+-   **Purpose**: Defines the `Uninstaller` interface for removing Drasi from an environment.
+-   **Key Components**:
+    -   `Uninstaller` (interface): The core interface with the `Uninstall(...)` method.
+    -   `MakeUninstaller(...)` (factory function): A factory that returns the correct `Uninstaller` implementation based on the target environment.
+
+## Kubernetes Implementation
+
+### `kubernetes_installer.go`
+
+-   **Purpose**: Provides the concrete implementation of the `Installer` interface for Kubernetes. This is a complex file that handles the multi-step process of deploying Drasi.
+-   **Key Responsibilities**:
+    -   Checking for and installing Dapr using a Helm chart.
+    -   Creating the target Kubernetes namespace.
+    -   Applying Kubernetes manifests for core infrastructure (Redis, Mongo), observability tools (Tempo, Prometheus, Grafana), and the Drasi control plane services.
+    -   Templating manifest files with correct image tags and registry paths.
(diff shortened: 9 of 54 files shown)


cli/cmd/ingress.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+
+	"drasi.io/cli/output"
+	"drasi.io/cli/sdk"
+	"drasi.io/cli/sdk/ingress"
+	"drasi.io/cli/sdk/registry"
+	"github.com/spf13/cobra"
+)
+
+func NewIngressCommand() *cobra.Command {
```
danielgerlag, line 28: How would the concept of "ingress" map to other platforms besides Kubernetes? eg. what would this look like for something like Azure Container Apps, etc.?

cli/cmd/ingress.go
```diff
@@ -0,0 +1,101 @@
...
+	var ingressCommand = &cobra.Command{
+		Use:   "ingress",
+		Short: "Manage Contour ingress for Drasi Sources and Reactions",
```
danielgerlag: We are hard coding for Contour?
Should we leave the installation / choice of a specific controller to the cluster admin?

cli/sdk/api_client.go
```diff
@@ -251,6 +259,91 @@ func (t *ApiClient) Watch(kind string, name string, output chan map[string]inter
...
+		if hasExternalEndpoint {
+			break
+		}
+	}
+
+	if hasExternalEndpoint {
+		// Get the Contour LoadBalancer IP
+		ip := t.getContourExternalIP()
+		ingressUrl := fmt.Sprintf("http://%s.drasi.%s.nip.io", reactionName, ip)
+		fmt.Printf("Ingress URL: %s\n", ingressUrl)
```
danielgerlag: Could we make use of the `output` service instead of directly printing to stdout?

cli/sdk/api_client.go
```diff
@@ -251,6 +259,91 @@ func (t *ApiClient) Watch(kind string, name string, output chan map[string]inter
...
+	}
+}
+
+func (t *ApiClient) getContourExternalIP() string {
```
danielgerlag: What if they are using a different ingress controller?

cli/sdk/ingress/contour_installer.go
```diff
@@ -0,0 +1,249 @@
...
+	defer tmpFile.Close()
+
+	// Set file permissions to 0600 (read/write for user only)
+	if err := os.Chmod(tmpFile.Name(), 0600); err != nil {
+		return "", err
+	}
+
+	tmpFile.Write(configBytes)
+
+	return tmpFile.Name(), nil
```
danielgerlag, line 268: Are there security risks to leaving the Kubernetes context, with auth certificates in a temporary file location?

control-planes/kubernetes_provider/src/controller/reconciler.rs
```diff
@@ -421,6 +442,88 @@ impl ResourceReconciler {
         Ok(())
     }
 
+    async fn get_contour_external_ip(&self) -> Option<String> {
```
danielgerlag: I would think that some organizations would have policies about what ingress controllers can be used. There are also Azure and AWS load balancer controllers which might be an obvious choice for some. How can we support any ingress controller instead of forcing Contour?

control-planes/kubernetes_provider/src/spec_builder/reaction.rs
```diff
@@ -136,7 +140,68 @@ impl SpecBuilder<ReactionSpec> for ReactionSpecBuilder {
...
+                                ..Default::default()
+                            };
+                            // Use just endpoint_name as key - ResourceReconciler will add reaction.id prefix
+                            let service_key = format!("{}-{}", service_name, endpoint_name);
+                            k8s_services.insert(service_key.clone(), service_spec);
+
+                            // Create Ingress resource with hostname-based routing
+                            let ingress_name = format!("{}-reaction-ingress", reaction.id);
+                            let mut annotations = BTreeMap::new();
+                            annotations.insert("kubernetes.io/ingress.class".to_string(), "contour".to_string());
```
danielgerlag: Can we make it extensible to support any ingress controller instead of hardcoding to contour?

ruokun-niu: Good point. I'll group all of these comments that are related to hardcoding to Contour and work on them together.

cli/sdk/ingress/contour_installer.go
ruokun-niu: This kubecontext file will be removed when the install completes
https://github.com/ruokun-niu/drasi-platform/blob/contour-ingress-controller/cli/sdk/ingress/contour_installer.go#L107

control-planes/kubernetes_provider/src/spec_builder/reaction.rs
ruokun-niu: @danielgerlag I have added this feature:

If the user does not want to install Contour and instead has another ingress controller installed, he can specify the ingress controller using a combination of three CLI flags: `--service` (the Kubernetes service of the ingress controller), `--ingress-namespace` (the amespace that the controller is installed in), and `--ingress-class-name` (name of the ingress class to use).

For instance, let's say I have `nginx` configured in my cluster and I dont want to use `contour`, I can use the following command:
```bash
drasi ingress init --use-existing --service ingress-nginx-controller --ingress-namespace ingress-nginx --ingress-class-name nginx
````

**Comment by ruokun-niu** [comment]
> Would it be possible to enhance the k3s embedded Docker installer to auto configure Drasi to use the Traefik ingress that comes pre-installed with k3s?

I can take a look

File: cli/cmd/ingress.go [hunk 8]
```diff
@@ -0,0 +1,286 @@
...
+	cmd.Flags().BoolVar(&useExisting, "use-existing", false, "Use existing ingress controller instead of installing Contour")
+	cmd.Flags().StringVar(&ingressServiceName, "ingress-service-name", "", "Name of the existing ingress controller LoadBalancer service (required with --use-existing)")
+	cmd.Flags().StringVar(&ingressNamespace, "ingress-namespace", "", "Namespace where the existing ingress controller is installed (required with --use-existing)")
+	cmd.Flags().StringVar(&ingressClassName, "ingress-class-name", "", "IngressClassName to use in ingress resources (required with --use-existing)")
+
+	return cmd
+}
+
+// UpdateIngressConfig updates the drasi-config ConfigMap with ingress controller configuration
+func UpdateIngressConfig(platformClient *sdk.KubernetesPlatformClient, drasiNamespace string, ingressClassName, ingressService, ingressNamespace string, output output.TaskOutput) error {
```
danielgerlag: Let's make this a method on the `PlatformClient` interface that is then implemented by the `KubernetesPlatformClient`, that way we can encapsulate all the logic for that platform type.

cli/sdk/ingress/contour_installer.go
```diff
@@ -0,0 +1,249 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	contourHelmRepoURL      = "https://charts.bitnami.com/bitnami"
+	contourChartName        = "contour"
+	contourDefaultNamespace = "projectcontour"
+)
+
+type ContourInstaller struct {
```
danielgerlag, line 44: We could abstract the concept of an ingress installer behind an interface and then have a concrete implementation that implements contour, kind of like how we have `PlatformClient` interface and `KubernetesPlatformClient` implementation.

cli/cmd/init.go
```diff
@@ -191,3 +198,38 @@ Usage examples:
 
 	return initCommand
 }
+
+func configureTraefikForDocker(namespace string, output output.TaskOutput) error {
```
danielgerlag: Could we encapsulate this inside the `DockerizedDeployer`?

cli/installers/dockerized_deployer.go
```diff
@@ -475,6 +479,9 @@ func (t *DockerizedDeployer) mergeDockerKubeConfig(name string, kubeconfig []byt
 		Namespace: "default",
 	}
 
+	// Set the new context as the current context
+	existingConfig.CurrentContext = contextName
```
danielgerlag: How did this currently happen?

cli/installers/dockerized_deployer.go
```diff
@@ -506,3 +513,169 @@ func pullImage(ctx context.Context, cli *client.Client, imageName string, output
...
+	}
+
+	output.InfoMessage("Successfully configured Traefik ingress for k3s Docker environment")
+	output.InfoMessage("Drasi Sources and Reactions with External endpoints will now be accessible via Traefik")
+
+	return nil
+}
+
+// updateIngressConfig updates the drasi-config ConfigMap with ingress controller configuration
+func updateIngressConfig(platformClient *sdk.KubernetesPlatformClient, drasiNamespace string, ingressClassName, ingressService, ingressNamespace string, output output.TaskOutput) error {
```
danielgerlag: Could you just call `UpdateIngressConfig` on the`KubernetesPlatformClient`?

cli/sdk/platform_client.go
```diff
@@ -31,6 +34,10 @@ type PlatformClient interface {
 	CreateTunnel(resourceType string, resourceName string, localPort uint16) error
 	SetSecret(name string, key string, value []byte) error
 	DeleteSecret(name string, key string) error
+
+	// Ingress management methods
+	UpdateIngressConfig(drasiNamespace string, ingressClassName, ingressService, ingressNamespace, gatewayIPAddress string, output output.TaskOutput) error
+	UpdateClusterRolePermissions(output output.TaskOutput) error
```
danielgerlag: Is `UpdateClusterRolePermissions` specific to ingress?
Why do we need a public function for it outside of `UpdateIngressConfig`?

cli/sdk/api_client.go
```diff
@@ -251,6 +259,110 @@ func (t *ApiClient) Watch(kind string, name string, output chan map[string]inter
...
+	if hasExternalEndpoint {
+		// Get the ingress controller LoadBalancer IP
+		ip := t.getIngressExternalIP()
+		ingressUrl := fmt.Sprintf("http://%s.drasi.%s.nip.io", reactionName, ip)
+		output.InfoMessage(fmt.Sprintf("Ingress URL: %s", ingressUrl))
+	}
+}
+
+func (t *ApiClient) getIngressExternalIP() string {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
```
danielgerlag: This is only going to work if your current kube context matches your current drasi instance, which may not always be the case.

cli/sdk/api_client.go
```diff
@@ -251,6 +259,110 @@ func (t *ApiClient) Watch(kind string, name string, output chan map[string]inter
...
+		return ""
+	}
+
+	// Create Kubernetes client
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return ""
+	}
+
+	configMap, err := clientset.CoreV1().ConfigMaps("drasi-system").Get(context.Background(), "drasi-config", metav1.GetOptions{})
```
danielgerlag: It feels like this functionality belongs on the platform client?

cli/sdk/platform_client.go
```diff
@@ -31,6 +34,10 @@ type PlatformClient interface {
 	CreateTunnel(resourceType string, resourceName string, localPort uint16) error
 	SetSecret(name string, key string, value []byte) error
 	DeleteSecret(name string, key string) error
+
+	// Ingress management methods
+	UpdateIngressConfig(drasiNamespace string, ingressClassName, ingressService, ingressNamespace, gatewayIPAddress string, output output.TaskOutput) error
```
danielgerlag: Would the platform client not already know the drasi namespace?
Could we bundle the other parameters into the `IngressConfig` struct?

cli/sdk/platform_client.go
```diff
@@ -340,3 +347,169 @@ func (t *KubernetesPlatformClient) getResourcePod(resourceType string, resourceN
...
 }
+
+// UpdateIngressConfig updates the drasi-config ConfigMap with ingress controller configuration
+// For regular ingress controllers, provide ingressService and ingressNamespace, leave gatewayIPAddress empty
+// For AGIC, provide ingressClassName and gatewayIPAddress, leave ingressService and ingressNamespace empty
+func (k *KubernetesPlatformClient) UpdateIngressConfig(drasiNamespace string, ingressClassName, ingressService, ingressNamespace, gatewayIPAddress string, output output.TaskOutput) error {
+	var taskName string
+	if gatewayIPAddress != "" {
+		taskName = "AGIC-Config"
+		output.AddTask(taskName, "Configuring Drasi for Azure Application Gateway Ingress Controller")
```
danielgerlag: Is there a way we could abstract out specific implementations (AGIC)?

ruokun-niu: I have updated locally, will push this change soon once I have something else figured out

cli/sdk/api_client.go
ruokun-niu: Good point. Let me see if there are any other ways to obtain the ingress address

cli/installers/dockerized_deployer.go
ruokun-niu: Maybe we can revert this if we cna figure out another way to get the ingress URLs.

cli/sdk/platform_client.go
ruokun-niu: This is for granting additional permission to list the services in any namespace, which is needed for retrieving the Ip address from the ingress controller. 
I think if we can get the ip address without queryig the ingress controller service, then we can remove this function.

seems like if we can improve the way to query the ingress ip address, we would be able to resolve a lot of the issues.

cli/sdk/api_client.go
ruokun-niu: Updated

cli/installers/dockerized_deployer.go
ruokun-niu: Reverted

cli/sdk/platform_client.go
ruokun-niu: We actually dont need this function anymore. I have deleted it.

cli/sdk/api_client.go
ruokun-niu: @danielgerlag  I have updated the CLI to use `KubernetesPlatformClient` to fetch the necessary information. This removes the need for us to create a new kube client from the kube context, as we can reuse the one in `KubernetesPlatformClient` (which uses the same kube context as the one in the Drasi instance).

cli/sdk/platform_client.go
ruokun-niu: Updated

cli/cmd/list.go
```diff
@@ -107,6 +107,16 @@ Usage examples:
 						}
 					}
 				}
+
+				// Add ingress URL for reactions and sources
+				if args[0] == "reaction" || args[0] == "source" {
+					ingressURL := platformClient.GetIngressURL(result[i].Id)
+					if ingressURL != "" {
+						item["ingress_url"] = ingressURL
+						statusFields["ingress_url"] = nil
```
danielgerlag: Is this something that should be coming back from the API in the list response?

control-planes/kubernetes_provider/src/controller/reconciler.rs
```diff
@@ -585,7 +585,12 @@ impl ResourceReconciler {
                         for rule in rules {
                             if let Some(host) = &rule.host {
                                 if host.contains("PLACEHOLDER") {
-                                    rule.host = Some(host.replace("PLACEHOLDER", &ip_suffix));
+                                    // For ALB, remove hostname entirely - let ALB generate its own URL
+                                    if ingress_class_name == "alb" {
```
danielgerlag: is there a cleaner way of doing this, where we don't hardcode specific logic for specific ingress controllers?

cli/cmd/list.go
ruokun-niu: Updated

control-planes/kubernetes_provider/src/controller/reconciler.rs
ruokun-niu: I have created a `IngressControllerConfig` struct, which updates the configuration based on the user input (ingress class name)

control-planes/kubernetes_provider/src/spec_builder/reaction.rs
```diff
@@ -136,7 +143,88 @@ impl SpecBuilder<ReactionSpec> for ReactionSpecBuilder {
...
+                            // Use just endpoint_name as key - ResourceReconciler will add reaction.id prefix
+                            let service_key = format!("{}-{}", service_name, endpoint_name);
+                            k8s_services.insert(service_key.clone(), service_spec);
+
+                            // Create Ingress resource with hostname-based routing
+                            let ingress_name = format!("{}-reaction-ingress", reaction.id);
+                            let mut annotations = BTreeMap::new();
+
+                            // Add ALB-specific annotations if using AWS Load Balancer Controller
+                            if runtime_config.ingress_class_name == "alb" {
```
danielgerlag: Could these annotations be part of the implementation config? Then we don't need to hardcode logic for "alb".

ruokun-niu: I think we should allow users to supply ingress-related annotations via the CLI during the init. I have this implemented but just need to figure out one more thing before submitting

ruokun-niu: I have added a new flag called `--ingress-annotation` that allow users to provide any annotation they want to add to their ingresses. Users can supply multiple annotations using their flag, and it works with Contour (installed via the CLI) or any other deployed ingress controller

cli/cmd/ingress.go
ruokun-niu: Resolving as we have discussed this offline during a sync

cli/cmd/ingress.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2024 The Drasi Authors.
```
ruokun-niu: 2025

ruokun-niu: @copilot can you change this to 2025

control-planes/kubernetes_provider/src/models.rs
```diff
@@ -100,6 +103,57 @@ pub struct RuntimeConfig {
...
+}
+
+#[derive(Clone, Debug)]
+pub struct IngressControllerConfig {
+    pub class_name: String,
+    pub supports_hostname: bool,
+    pub path_type: String,
+}
+
+impl IngressControllerConfig {
```
ruokun-niu: Note: read from config

e2e-tests/fixtures/ingress-fixture.js
```diff
@@ -0,0 +1,247 @@
...
+   * @param {string} ingressServiceName - Name of the ingress service (default: contour-envoy)
+   * @param {string} ingressNamespace - Namespace of the ingress service (default: projectcontour)
+   */
+  constructor(reactionName, queryIds, ingressServiceName = 'contour-envoy', ingressNamespace = 'projectcontour') {
+    this.reactionName = reactionName;
+    this.queryIds = queryIds;
+    this.ingressServiceName = ingressServiceName;
+    this.ingressNamespace = ingressNamespace;
+    
+    this.portForward = new PortForward(ingressServiceName, 80, ingressNamespace);
```
danielgerlag: Is the point of using ingress to not use a port forward?

e2e-tests/fixtures/ingress-fixture.js
```diff
@@ -0,0 +1,247 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const PortForward = require('./port-forward');
+const axios = require('axios');
+const signalR = require("@microsoft/signalr");
+
+class IngressFixture {
```
danielgerlag, line 21: This is still a SignalR fixture, but perhaps it should to a flag if it should use port forwarding or an ingress?

e2e-tests/04-kubernetes-scenario/resources.yaml
```diff
@@ -172,3 +172,17 @@ spec:
...
+name: reaction1
+spec:
+  kind: SignalR
+  queries:
+    risky-containers:
+  services:
+    reaction:
+      endpoints:
+        service:
+          setting: external
```
danielgerlag, line 187: Should external be the default for SignalR?

ruokun-niu: Good point. Setting this to default will cause an ingress and a service to both be created.

ruokun-niu: I'm unsure whether the resource provider will throw an error if this is set to 'external' by default and the user does not have an ingress controller configured. Let me verify this.

e2e-tests/fixtures/ingress-fixture.js
ruokun-niu: Yes, however in this case we are port-forwarding the ingress controller (contour) in this case. Different from the signalr Fixture, we are testing to see if

1. an ingress has been created
2. if the ingress directs to the service properly

ruokun-niu: Actually, it might be possible to do this without port-forwarding if we configure our kind cluster properly using a config file. Let me give this a try

e2e-tests/04-kubernetes-scenario/resources.yaml
ruokun-niu: Just tested this. No error was thrown. I have updated the reaction-provider file

e2e-tests/fixtures/ingress-fixture.js
ruokun-niu: Ok I was able to get it working without port-forwarding 🚀 . The e2e-test will create a cluster with a specific config (created a kind-config.yaml in the `e2e-tests` folder), and the signalr hub can be reached at `http://localhost/hub`

e2e-tests/fixtures/ingress-fixture.js
```diff
@@ -58,12 +55,12 @@ class IngressFixture {
...
     await this.signalr?.stop();
-    this.portForward?.stop();
+    // No port forward to stop
   }
   
   async connectSignalR() {
-    // Create SignalR connection through ingress
-    const hubUrl = `http://127.0.0.1:${this.localPort}/hub`;
+    // Create SignalR connection through ingress via localhost
+    const hubUrl = `http://localhost/hub`;
```
danielgerlag: Port 80 might not always be available in a dev environment, often devs have a web server running on that port.
I think we had something to detect a free port within a range?

ruokun-niu: Updated this. The test infra will first look for a free port, and update the `kind-config.yaml` to use this value. This port will also be used in `ingress-fixture`

danielgerlag (approved):

amansinghoriginal (approved):

PR #296, opened 2025-09-03
Fix alternate schema bug in sql proxy

danielgerlag:
This pull request refactors the way tables and schemas are handled in the SQL proxy by introducing a new `TableRegistry` class and updating the table cursor logic to support schema-qualified table names. The changes improve robustness when dealing with tables from multiple schemas and ensure that only registered tables are processed, and fixes a bug where tables in the non-default schema could not be bootstrapped.

**Schema and Table Registry Improvements:**

* Added a new `TableRegistry` class to manage table-to-schema mappings, parsing the `tables` config and providing lookup methods for table existence and schema name. (`sources/relational/sql-proxy/src/main/java/io/drasi/TableRegistry.java`)
* Updated `ResultStream` to instantiate and use `TableRegistry`, checking if each requested table exists and retrieving its schema before creating `TableCursor` instances. Unregistered tables are logged and skipped. (`sources/relational/sql-proxy/src/main/java/io/drasi/ResultStream.java`) [[1]](diffhunk://#diff-3bf6025a2a9d7303bbc3bb129f2e41a5c7e69067021c3fd64eee65cbc3fabc8bR17-R23) [[2]](diffhunk://#diff-3bf6025a2a9d7303bbc3bb129f2e41a5c7e69067021c3fd64eee65cbc3fabc8bL30-R37)

**Table Cursor and Query Handling:**

* Refactored `TableCursor` to accept both `schemaName` and `tableName` in its constructor, and to use these values when building SQL queries and reading table mappings. (`sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java`)
* Updated SQL query construction in `TableCursor` to use fully qualified table names (including schema when present), ensuring correct querying in multi-schema environments. (`sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java`)
* Simplified and clarified the logic for reading table mappings and primary keys, passing schema and table names directly and improving error messages for missing primary keys. (`sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java`)

### DIFF
--- a/sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java
+++ b/sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java
@@ -16,24 +16,27 @@ import java.util.Collections;
 
 class TableCursor {
     private static final Logger log = LoggerFactory.getLogger(TableCursor.class);
-    public final String tableName;
+    public final String schemaName;
+    public final String tableName;    
     public NodeMapping mapping;
     public ResultSet resultSet;
     public ResultSetMetaData metaData;
     public int columnCount;
 
-    public TableCursor(String tableName) {
+    public TableCursor(String schemaName, String tableName) {
+        this.schemaName = schemaName;
         this.tableName = tableName;
     }
 
     private void Init(Connection connection) throws SQLException {
         if (resultSet == null) {
-            mapping = ReadMappingFromSchema(tableName, connection);
+            mapping = ReadMappingFromSchema(schemaName, tableName, connection);
             var statement = connection.createStatement();
 
             String quote = SourceProxy.GetConfigValue("connector").equalsIgnoreCase("MySQL") ? "`" : "\"";
             var sanitizedTableName = tableName.replace(quote, "").replace(";", "");
-            resultSet = statement.executeQuery("SELECT * FROM " + quote + sanitizedTableName + quote);
+            var fullyQualifiedTableName = schemaName != null ? quote + schemaName + quote + "." + quote + sanitizedTableName + quote : quote + sanitizedTableName + quote;
+            resultSet = statement.executeQuery("SELECT * FROM " + fullyQualifiedTableName);
             metaData = resultSet.getMetaData();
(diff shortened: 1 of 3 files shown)


sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java
```diff
@@ -81,23 +84,15 @@ private String SanitizeNodeId(String nodeId) {
...
-        if (table.contains(".")) {
-            var tableComps = table.split("\\.");
-            schemaName = tableComps[0];
-            tableName = tableComps[1];
-        }
 
         var rs = metadata.getPrimaryKeys(null, schemaName, tableName);
         if (!rs.next())
-            throw new SQLException("No primary key found for " + table);
+            throw new SQLException("No primary key found for " + (schemaName != null ? schemaName + "." + tableName : tableName));
```
amansinghoriginal, line 92: Should we make this into a function `schemaName != null ? schemaName + "." + tableName : tableName` ? Since it is repeated a couple of times.

amansinghoriginal (approved):
LGTM

sources/relational/sql-proxy/src/main/java/io/drasi/TableCursor.java
```diff
@@ -81,23 +84,15 @@ private String SanitizeNodeId(String nodeId) {
...
         var mapping = new NodeMapping();
-        mapping.tableName = rs.getString("TABLE_SCHEM") + "." + tableName;
+        String actualSchema = rs.getString("TABLE_SCHEM");
```
agentofreality, line 94: Is TABLE_SCHEM a typo that we are perpetuating?

danielgerlag: `TABLE_SCHEM` is part of the JDBC drivers, for returning metadata.
https://learn.microsoft.com/en-us/sql/connect/jdbc/reference/getschemas-method?view=sql-server-ver17#remarks

PR #297, opened 2025-09-05
Image validation workflow

ruokun-niu:
This pull request introduces a new image validation workflow into the release process to ensure that all Drasi container images meet architecture requirements before a release is finalized. The workflow checks for the presence of required architectures (amd64 and arm64) for each image, tests pulling images on real hardware runners, and integrates these checks into the existing release pipeline.

**Release process improvements:**

* Added a new `validate-images` job to the `.github/workflows/draft-release.yml` workflow, which runs after manifest creation and before the final release, ensuring image validation is a required step for releases. [[1]](diffhunk://#diff-3d80a0bd2aa867253847d98ff0fddb8982d61587bee3ddfb9906f4aafbe55329R345-R352) [[2]](diffhunk://#diff-3d80a0bd2aa867253847d98ff0fddb8982d61587bee3ddfb9906f4aafbe55329R364)

**Image validation workflow:**

* Introduced a new workflow file, `.github/workflows/image-validation.yml`, which:
  - Validates that all images for the specified release tag exist and support the required architectures (most images require both amd64 and arm64; `reaction-dataverse` only requires amd64).
  - Repeats the validation for "azure-linux" image variants.
  - Tests pulling all images on both amd64 and arm64 runners to confirm images can be pulled successfully on each architecture.
  - Fails the release if any images are missing, lack required architecture support, or cannot be pulled on the appropriate hardware.

Fixes: #issue_number


.github/workflows/image-validation.yml
```diff
@@ -0,0 +1,383 @@
...
+            echo "Requirements:"
+            echo "- Most images: amd64 + arm64 support required"
+            echo "- reaction-dataverse: amd64 support only required"
+            exit 1
+          fi
+          
+          echo ""
+          echo "✅ SUCCESS: All images meet their architecture requirements for version $IMAGE_TAG"
+
+      - name: Validate azure-linux variant architecture support
```
amansinghoriginal: Would a matrix strategy be better to reduce the workflow file size?
Something like:
```
   multi-arch-validation:
...
     strategy:
       matrix:
         variant: ["", "-azure-linux"]
     steps:
       - name: Validate ${{ matrix.variant || 'default' }} variant architecture support
         run: |
          IMAGE_TAG="${{ needs.get-version.outputs.image_tag }}${{ matrix.variant }}"
```

.github/workflows/image-validation.yml
```diff
@@ -0,0 +1,383 @@
...
+          
+          # Images that only need amd64 support
+          AMD64_ONLY_IMAGES="ghcr.io/drasi-project/reaction-dataverse"
+          
+          failed_pulls=()
+          skipped_images=()
+          
+          for image in $DRASI_IMAGES; do
+            # Skip reaction-dataverse unless we're on x86-64 runner
+            if echo "$AMD64_ONLY_IMAGES" | grep -q "$image" && [[ "${{ matrix.runner }}" != "oracle-vm-8cpu-32gb-x86-64" ]]; then
```
amansinghoriginal: Should we check like `[[ "${{ matrix.arch }}" == "arm64" ]]` instead of the runner name?

amansinghoriginal (approved):
Approving with suggestions.

.github/workflows/image-validation.yml
ruokun-niu: Good point. Will update this

danielgerlag: Can it make use of the existing matrix?

.github/workflows/image-validation.yml
```diff
@@ -0,0 +1,383 @@
...
+          echo ""
+          echo "✅ SUCCESS: All images meet their architecture requirements for version $IMAGE_TAG"
+
+      - name: Validate azure-linux variant architecture support
+        run: |
+          echo "🔍 Validating azure-linux variant multi-architecture support..."
+          echo "📋 Using image tag: ${{ needs.get-version.outputs.image_tag }}-azure-linux"
+          
+          # Images that only need amd64 support
+          AMD64_ONLY_IMAGES="ghcr.io/drasi-project/reaction-dataverse"
```
danielgerlag: This could come from the matrix too

.github/workflows/image-validation.yml
ruokun-niu: Good suggestion

.github/workflows/image-validation.yml
```diff
@@ -0,0 +1,262 @@
...
+    runs-on: ${{ matrix.runner }}
+    needs: [get-version, multi-arch-validation]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runner: oracle-vm-8cpu-32gb-x86-64
+            variant: ""
+            amd64_only_images: "ghcr.io/drasi-project/reaction-dataverse"
```
danielgerlag: Why are we only doing an amd64 build for dataverse?

ruokun-niu: The Dataverse reaction is only built with amd64 architecture in the release workflow. IIRC, we had a dependency that would only work when we build the image in amd architecture.

danielgerlag: Looking at the dependencies, there is nothing that stands out as being an issue. Can we confirm this?

ruokun-niu: Yep let me check again

ruokun-niu: I've added the arm64 variant of the dataverse reaction to both the release workflow and the image validation workflow

.github/workflows/image-validation.yml
```diff
@@ -0,0 +1,225 @@
...
+  image-pull-test:
+    name: Image Pull Test
+    runs-on: ${{ matrix.runner }}
+    needs: [get-version, multi-arch-validation]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runner: oracle-vm-8cpu-32gb-x86-64
```
danielgerlag: If we are just pulling images, do we need an 8cpu VM?

ruokun-niu: probably not, let me update this

danielgerlag (approved):

PR #309, opened 2025-09-19
Add automated OpenAPI specification for `mgmt_api`

amansinghoriginal:
# Description

This pull request introduces a robust, automated system for generating and maintaining an OpenAPI 3.0 specification for the Drasi Management API.

The primary goal is to provide a definitive, version-controlled API contract that is always in sync with the implementation. This enhances developer experience, improves documentation, and enables automated tooling for API validation and client generation.

This work is the result of the strategy outlined in `.agents/api_spec_guidelines.md`.

### 1. OpenAPI Specification Generation
- **`utoipa` Integration:** The `utoipa` crate has been integrated into the `mgmt_api` service to generate the OpenAPI spec directly from the Rust code.
- **Route Refactoring & Macro Removal:** The core of this change involved a significant refactoring within the `mgmt_api` service. The previous route-generation macros (`v1_crud_api!`, `v1_crud_api_provider!`) have been completely removed. They are replaced with:
    - New, dedicated modules for each resource (e.g., `sources.rs`, `reactions.rs`).
    - Explicit Actix handler functions (`upsert`, `get`, `list`, etc.) within each module.
    - `#[utoipa::path]` annotations on each handler function, which enables the automatic spec generation.
- **Generated Spec:** The resulting `openapi.yaml` is now checked into the repository at `control-planes/mgmt_api/openapi.yaml`.

### 2. Automation and CI/CD Integration
- **Pre-Commit Hook:** A new Git pre-commit hook (`.githooks/pre-commit`) has been added. It automatically regenerates the OpenAPI spec before each commit and fails if the generated spec has changed, forcing the developer to include the updated spec.
- **CI Verification:** The `build-test.yml` GitHub Actions workflow has been updated with two new jobs:
    - `verify-openapi-spec`: Ensures the committed spec is never out of sync with the code.
    - `detect-breaking-api-changes`: Uses `openapi-diff` to prevent backward-incompatible changes from being merged into `main`, thus enforcing the `v1` API contract.

### 3. API Versioning Alignment
- **Package Version:** The `mgmt_api` crate version in `Cargo.toml` has been updated from `0.1.0` to `1.0.0`. This aligns the package's major version with the `/v1` API path, correctly signaling its stability.
- **Spec Version:** The version in the generated `openapi.yaml` is now correctly set to `1.0.0`.

### 4. Tooling and Refinements
- **Makefile Targets:** New `openapi-*` targets have been added to the `control-planes/mgmt_api/Makefile` for developers to easily generate, validate, and serve the spec locally.
- **Dead Code Removal:** The now-unused route-generation macros (`v1_crud_api!`) have been removed, cleaning up the codebase.
- **WebSocket Documentation:** A placeholder has been added to the OpenAPI spec to document the `/v1/debug` WebSocket endpoint.

## Type of change

This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Drasi.


.github/workflows/build-test.yml
```diff
@@ -33,6 +33,87 @@
...
+            echo "ERROR: The generated OpenAPI specification is out of date."
+            echo "Please run 'make -C control-planes/mgmt_api openapi-generate' and commit the changes."
+            exit 1
+          fi
+          echo "OpenAPI specification is up to date."
+
+  detect-breaking-api-changes:
+    name: Detect Breaking API Changes
+    runs-on: ubuntu-latest
+    # TODO: Enable this job after the initial OpenAPI PR is merged
```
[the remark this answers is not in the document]
amansinghoriginal: This is required because the step to detect breaking changes cannot run when we're bootstrapping a new spec, as if it was enabled, it will compare with an empty file and fail.

control-planes/mgmt_api/src/api/v1/debug.rs
```diff
@@ -0,0 +1,37 @@
...
+        content = inline(QuerySpecDto),
+        description = "Query specification sent via WebSocket after connection is established",
+        content_type = "application/json"
+    )
+)]
+#[allow(dead_code)]
+async fn debug_websocket_placeholder() {
+    // This function exists solely for OpenAPI documentation
+    // The actual WebSocket handler is implemented elsewhere
+    unimplemented!("This is a documentation placeholder");
```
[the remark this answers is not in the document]
amansinghoriginal: This is for documentation purposes only.

.githooks/pre-commit
```diff
@@ -7,6 +7,22 @@ case ":${PATH}:" in
         ;;
 esac
 
+echo "Running pre-commit hook: Checking for API spec changes..."
```
danielgerlag, line 10: In my experience, pre-commit hooks are not reliable, as devs often don't set execution permissions or there are cross platform issues.

control-planes/mgmt_api/src/api/v1/continuous_queries.rs
```diff
@@ -0,0 +1,213 @@
...
+#[utoipa::path(
+    put,
+    path = "/v1/continuousQueries/{id}",
+    tag = "Continuous Queries",
+    params(
+        ("id" = String, Path, description = "Query ID")
+    ),
+    request_body = QuerySpecDto,
+    responses(
+        (status = 200, description = "Query created or updated", body = ContinuousQueryDto),
```
danielgerlag: We don't support updating queries

control-planes/mgmt_api/src/api/v1/continuous_queries.rs
```diff
@@ -0,0 +1,213 @@
...
+    params(
+        ("id" = String, Path, description = "Query ID")
+    ),
+    responses(
+        (status = 200, description = "Stream of query results as newline-delimited JSON (NDJSON). Each line contains a ResultEventDto object.", body = [ResultEventDto], content_type = "application/x-ndjson"),
+        (status = 404, description = "Query not found"),
+        (status = 500, description = "Internal server error")
+    )
+)]
+pub async fn watch(service: web::Data<DebugService>, id: web::Path<String>) -> impl Responder {
```
danielgerlag, line 207: I don't think we have any of the E2E tests that hit the `watch` or `debug` endpoints. We'll need to test these.

control-planes/mgmt_api/src/api/v1/models/result.rs
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::Serialize;
+use serde_json::{Map, Value};
+use utoipa::ToSchema;
+
+#[derive(Serialize, ToSchema)]
+#[serde(tag = "kind")]
+pub enum ResultEventDto {
```
danielgerlag, line 21: Sine the ResultEvent originates from the QueryHost, would it make sense for the QueryHost to be the source of truth for that schema and publish an OpenAPI spec that the Management API consumes?

control-planes/mgmt_api/src/api/v1/openapi.rs
```diff
@@ -0,0 +1,152 @@
...
+        (name = "Sources", description = "Source resource management"),
+        (name = "Query Containers", description = "Query Container resource management"),
+        (name = "Reactions", description = "Reaction resource management"),
+        (name = "Continuous Queries", description = "Continuous Query resource management"),
+        (name = "Source Providers", description = "Source Provider registration and management"),
+        (name = "Reaction Providers", description = "Reaction Provider registration and management"),
+    ),
+    info(
+        title = "Drasi Management API",
+        version = "1.0.0",
```
danielgerlag, line 137: Should we harmonize thus version with our release versions?

control-planes/mgmt_api/src/api/v1/query_containers.rs
```diff
@@ -0,0 +1,171 @@
...
+    )
+)]
+pub async fn ready_wait(
+    service: web::Data<QueryContainerDomainService>,
+    id: web::Path<String>,
+    params: web::Query<ReadyWaitParams>,
+) -> impl Responder {
+    log::debug!("readywait_resource: {:?}", id);
+
+    if params.timeout > 300 {
```
danielgerlag: I see this timeout in multiple places, should we manage it from a config file or constant?

control-planes/mgmt_api/src/api/v1/models/result.rs
danielgerlag: EDIT: we have a TypeSpec definition here: https://github.com/drasi-project/drasi-platform/blob/main/typespec/query-output/main.tsp

control-planes/mgmt_api/src/api/v1/query_containers.rs
amansinghoriginal: Fixed

control-planes/mgmt_api/src/api/v1/openapi.rs
amansinghoriginal: API version will change when we make API contract changes, where as we do platform releases for any features, new source/reaction, bug-fixes.
If we couple them, we will bump API version for changes that don't affect the API at all.

Even platforms like Kubernetes keep API version and platform version separate right?

control-planes/mgmt_api/src/api/v1/models/result.rs
amansinghoriginal: For utopia to generate the OpenAPI spec, we need Rust DTOs with #[derive(ToSchema)] 
This spec will be consumed by Go CLI later (openAPI -> codegen -> go client)

TypeSpec -> quicktype doesn't support `#[derive(ToSchema)]`.
Resource management DTO sare not in typespec yet anyways.

amansinghoriginal: I tried exploring Typespec's OpenAPI emitter instead of utopia, as that would give us a single source of truth. But, it requires custom templates for discriminated unions, and more complex build pipeline.

.githooks/pre-commit
amansinghoriginal: Agreed, but that's why we have the `verify-openapi-spec` job in `build-test.yml` workflow. This will fail the build if a PR changes cause the openapi.yaml to be out of sync. So this will ensure that no outdated spec can be merged, regardless of whether the local hook runs successfully.

We also have `detect-breaking-api-changes` job that uses openapi-diff to prevent merging backward incompatible changes.

control-planes/mgmt_api/src/api/v1/continuous_queries.rs
amansinghoriginal: Fixed. Now only supporting Create, not upsert.

control-planes/mgmt_api/src/api/v1/source_providers.rs
```diff
@@ -0,0 +1,134 @@
...
+#[utoipa::path(
+    put,
+    path = "/v1/sourceProviders/{id}",
+    tag = "Source Providers",
+    params(
+        ("id" = String, Path, description = "Source Provider ID")
+    ),
+    request_body = ProviderSpecDto,
+    responses(
+        (status = 200, description = "Source Provider created or updated", body = SourceProviderDto),
```
ruokun-niu, line 31: Do providers have a status?

control-planes/mgmt_api/src/api/v1/sources.rs
```diff
@@ -0,0 +1,169 @@
...
+    {
+        Ok(res) => match res {
+            true => HttpResponse::Ok().finish(),
+            false => HttpResponse::ServiceUnavailable().finish(),
+        },
+        Err(e) => e.into(),
+    }
+}
+
+pub fn configure_routes() -> actix_web::Scope {
```
ruokun-niu, line 166: What about `describe`?

amansinghoriginal: Doesn't `describe` command in the CLI also invokes the normal `GET` method on the HTTP resource like `source` or `continuousquery`.

control-planes/mgmt_api/src/api/v1/source_providers.rs
amansinghoriginal: Good question. I see the [definitions like this here](https://github.com/drasi-project/drasi-platform/blob/550627abf4d856896a69349e62ec7781383b5d36/control-planes/mgmt_api/src/domain/models.rs#L224). But they don't seem to be used anywhere?

control-planes/mgmt_api/src/api/v1/continuous_queries.rs
amansinghoriginal: Verified both manually.

control-planes/mgmt_api/src/api/v1/sources.rs
ruokun-niu: Right, that's a good point. Thanks for the reminder

control-planes/mgmt_api/src/api/v1/source_providers.rs
ruokun-niu: Yeah, I'm just not sure if it makes sense to include statuses for providers if we are not using it. I guess it's fine if we leave it there.

control-planes/mgmt_api/src/api/v1/openapi.rs
danielgerlag: How would the CLI know which version it supports / how would we release / version them together?

control-planes/mgmt_api/src/api/v1/debug.rs
```diff
@@ -0,0 +1,51 @@
...
+    request_body(
+        content = inline(QuerySpecDto),
+        description = "Query specification sent as WebSocket text message after connection is established (not as HTTP body)",
+        content_type = "application/json"
+    )
+)]
+#[allow(dead_code)]
+async fn debug_websocket_placeholder() {
+    // This function exists solely for OpenAPI documentation
+    // The actual WebSocket handler is implemented elsewhere
```
danielgerlag: where is it implemented?

.githooks/pre-commit
```diff
@@ -7,6 +7,22 @@ case ":${PATH}:" in
...
+
+# Generate the latest spec based on the current code
+make -C control-planes/mgmt_api openapi-generate
```
danielgerlag, line 13: We should add documentation for developers on how to configure these git hooks, I don't think they will just magically work after pulling the repo.

control-planes/mgmt_api/src/api/v1/debug.rs
amansinghoriginal: Actual implementation is in `control-planes/mgmt_api/src/api/v1/mod.rs`

amansinghoriginal: The debug.rs file exists solely for OpenAPI documentation because:
  1. OpenAPI 3.0 doesn't fully support WebSocket semantics
  2. The #[utoipa::path] macro needs a function to attach to
  3. The placeholder documents the WebSocket protocol for API consumers

control-planes/mgmt_api/src/api/v1/openapi.rs
amansinghoriginal: Clients (like the CLI) currently hardcode the version in the URL path (e.g. `v1`). This path needs to be changed only when breaking changes are made to the API. 

URL Path version will encode only the major version for client routing, but the OpenAPI Spec can have the full semantic version for contract tracking.

URL Path version will change only on breaking changes, where as sem-ver in API spec can keep incrementing based on the type of change:
- Patch (1.0.x): Bug fixes, doc changes
- Minor (1.x.y): New features, backward compatible
- Major (x.0.0): Breaking Changes (add new path for clients)

amansinghoriginal: We have a few options for what could be done in future for CLI to know what version to call.
1. Version Discovery endpoint (`GET /version`)
2. CLI can check compatibility before invoking the API endpoints.
3. CLI can warn users when they need to upgrade
4. Can add API version headers (Request api-version, Response api-version, etc.)

control-planes/mgmt_api/src/api/v1/debug.rs
amansinghoriginal: I am now moving it to the debug.rs, so that we have single source of truth - both implementation and docs in one place.

.githooks/pre-commit
amansinghoriginal: We already have it as part of our pre-requisites contributing docs.

https://github.com/drasi-project/drasi-platform/tree/main/docs/contributing/contributing-code/contributing-code-prerequisites

> Enable Git Hooks We use pre-commit hooks to catch some issues early, enable the local git hooks using the following commands
> 
> chmod +x .githooks/pre-commit
> git config core.hooksPath .githooks

PR #315, opened 2025-09-25
Update Debezium & Azure Identity libraries

danielgerlag:
This pull request introduces support for authenticating to Azure PostgreSQL databases using Azure Identity (Microsoft Entra Workload ID) in both the Debezium reactivator and SQL proxy components. It also updates several dependencies and adds logic to select authentication mode based on configuration.

**Azure Identity Integration:**

* Added `AzureIdentityTokenFetcher` and `AzureIdentityPostgreSQLConnection` classes to fetch Azure AD tokens and establish PostgreSQL connections using Azure Identity, enabling managed identity authentication. [[1]](diffhunk://#diff-17488154f74a9c9c4614289dee019dad449322e026dff132ad398654ca87e784R1-R25) [[2]](diffhunk://#diff-bbe7a51bf2df8216619cd2b1fc289b9e0c62dfe694d6abd3a200fa24fa850da6R1-R52)
* Updated the Debezium reactivator (`PostgreSql.java`) and SQL proxy (`ResultStream.java`) to use Azure Identity for authentication when `IDENTITY_TYPE` is set to `MicrosoftEntraWorkloadID`, including setting the token as the database password and enforcing SSL. [[1]](diffhunk://#diff-21809a017dce0750f11de9484fd2e8edc6f495d46f775631eb6bc6c8a687829eL107-R122) [[2]](diffhunk://#diff-3bf6025a2a9d7303bbc3bb129f2e41a5c7e69067021c3fd64eee65cbc3fabc8bR72-R78)

**Dependency Updates:**

* Upgraded `debezium` from `2.7.3.Final` to `3.2.2.Final` and `azure-identity` from `1.9.2` to `1.18.0` in relevant Maven files for compatibility with Azure Identity features. [[1]](diffhunk://#diff-c31c404751f54d4170da59c7237804068dfb5c8575e1eb6d72397993c9208fd5L33-R33) [[2]](diffhunk://#diff-c31c404751f54d4170da59c7237804068dfb5c8575e1eb6d72397993c9208fd5L72-R72) [[3]](diffhunk://#diff-e2ba3e04ed17cde245dc36e9a3c7fa6bab899664c69929dceadcb16ef6d0e731L63-R63)
* Updated PostgreSQL JDBC driver from `42.7.4` to `42.7.8` for the SQL proxy.

**Configuration Improvements:**

* Added error retry configuration (`errors.max.retries`) to the Debezium connector setup for improved reliability.

**YAML Spec Simplification:**

* Removed `password` and `user` fields from the default source provider spec, reflecting the move to token-based authentication for Azure.

### DIFF
--- a/cli/installers/resources/default-source-providers.yaml
+++ b/cli/installers/resources/default-source-providers.yaml
@@ -60,8 +60,6 @@ spec:
       - database
       - host
       - port
-      - password
-      - user
       - tables
 ---
 apiVersion: v1
--- a/sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/PostgreSql.java
+++ b/sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/PostgreSql.java
@@ -25,8 +25,8 @@ import io.drasi.DatabaseStrategy;
 import io.drasi.models.NodeMapping;
 import io.drasi.models.RelationalGraphMapping;
 import io.drasi.source.sdk.Reactivator;
+import io.drasi.source.sdk.SourceProxy;
 
-import org.checkerframework.checker.units.qual.t;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -90,7 +90,7 @@ public class PostgreSql implements DatabaseStrategy {
     @Override
     public Configuration createConnectorConfig(Configuration baseConfig) {
         var publicationSlotName = "rg_" + baseConfig.getString("name");
-        return Configuration.create()
+        var result = Configuration.create()
                 // Start with the base configuration.
                 .with(baseConfig)
                 // Specify the Postgres connector class.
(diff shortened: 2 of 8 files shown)


ruokun-niu (approved):

cli/installers/resources/default-source-providers.yaml
```diff
@@ -60,8 +60,6 @@ spec:
       - database
       - host
       - port
-      - password
```
amansinghoriginal, line 63: Should we make `user` and `password` optional here along with `identity` block, and make the `mgmt_api` enforce that atleast one of them is present?

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/PostgreSql.java
```diff
@@ -104,8 +104,22 @@ public Configuration createConnectorConfig(Configuration baseConfig) {
                 // Name of replication slot for streaming changes from the database. Default is debezium. 
                 .with("slot.name", publicationSlotName)
                 // If started first time, start from beginning, else start from last stored LSN.
-                .with("snapshot.mode", "no_data")
-            .build();
+                .with("snapshot.mode", "no_data");
+
+        String identityType = SourceProxy.GetConfigValue("IDENTITY_TYPE");
```
amansinghoriginal, line 109: Should we reconsider (maybe in a separate PR) adding fields to source-provider instead of relying on undocumented environment variables like `IDENTITY_TYPE` ?

Should the SDK provide a better way of achieving this? I worry that we're not using the schema-driven approach properly.

cli/installers/resources/default-source-providers.yaml
danielgerlag: I think we should avoid making global assumptions about how all sources should work in the Management API. Ideally we want to offload this validation to the implementor of each source.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/PostgreSql.java
danielgerlag: We have varying degrees of support for abstracting that environment variable across our SDKs, I think the .net version has this. We also do not yet have any documentation on building a source, I think that would be the place to document these. Could you clarify what you mean by "schema-driven approach"?

amansinghoriginal: By schema-driven I meant that source-provider defines the schema of the Source Config, so it can drive what fields  are supported in the config.

danielgerlag: The identity type is a higher level field in the YAML that is applicable to all sources / reactions. The config schema is set for the configuration section, which is specific per component.

PR #328, opened 2025-10-16
Bug fix: Add health check to signalr reaction

ruokun-niu:
# Description

**Issue:**
The SignalR deployment in the `drasi-system` namespace is experiencing 502 Bad Gateway errors when accessed through Azure Application Gateway Ingress Controller (AGIC).

### Root Cause

The issue stems from failed health probe validation:

1. **Health Probe Failure**: Application Gateway health probes targeting the default `/` path are receiving HTTP 404 responses from the backend service. The SignalR application only maps the `/hub` endpoint, causing the root path (`/`) to return 404 responses. This differs from the debug reaction, which handle the `/` path correctly and don't require custom health endpoints for Application Gateway health probes.
2. **Backend Pool Status**: As a result, the backend pool ( is being marked as unhealthy
3. **Gateway Response**: Per AGIC documentation, when backend pods are marked unhealthy, Application Gateway returns 502 Bad Gateway errors to clients

This behavior aligns with Azure Application Gateway's standard ingress controller pattern, where unhealthy backend pools trigger 502 responses rather than routing traffic to potentially failing services.

**Fix:**

This pull request introduces a small but important improvement to the SignalR reaction service. The main change is the addition of a health check endpoint to the application, which allows external systems to verify that the service is running and healthy.

Service observability:

* Added a `/health` HTTP GET endpoint to the SignalR service in `Program.cs`, returning a simple "Healthy" response for monitoring and health checks.
* Imported the necessary `Microsoft.AspNetCore.Http` namespace to support the new endpoint.

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -404,4 +404,11 @@ jobs:
           E2E_SYNC_VECTORSTORE_AZURE_OPENAI_EMBEDDING_MODEL: ${{ secrets.E2E_SYNC_VECTORSTORE_AZURE_OPENAI_EMBEDDING_MODEL }}
         run: npm test
 
-  
+      - name: Report vectorstore test status
+        if: always()
+        run: |
+          if [ -z "${{ secrets.E2E_SYNC_VECTORSTORE_AZURE_OPENAI_KEY }}" ]; then
+            echo "::notice::Vectorstore tests were skipped (no OpenAI credentials available). This is expected for fork-based PRs."
+          else
+            echo "::notice::Vectorstore tests were executed with Azure OpenAI credentials."
+          fi
--- a/e2e-tests/07-sync-inmemory-vectorstore-scenario/inmemory-vectorstore.test.js
+++ b/e2e-tests/07-sync-inmemory-vectorstore-scenario/inmemory-vectorstore.test.js
@@ -190,29 +190,39 @@ async function verifyDocumentDeletion(reactionName, queryId, expectedCount) {
 }
 
 beforeAll(async () => {
+  // Check if Azure OpenAI credentials are available
+  const azureKey = process.env.E2E_SYNC_VECTORSTORE_AZURE_OPENAI_KEY;
+  const azureEndpoint = process.env.E2E_SYNC_VECTORSTORE_AZURE_OPENAI_ENDPOINT;
+  const azureModel = process.env.E2E_SYNC_VECTORSTORE_AZURE_OPENAI_EMBEDDING_MODEL;
+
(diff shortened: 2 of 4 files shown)


reactions/signalr/signalr-reaction/Drasi.Reactions.SignalR/Program.cs
```diff
@@ -73,6 +74,7 @@ public static async Task Main(string[] args)
 
         hub.UseCors();
         hub.UseRouting();
+        hub.MapGet("/health", () => Results.Ok("Healthy"));
```
danielgerlag: Should we return unhealthy if there is a problem, for example the upstream azure signalr service cannot be reached, or we can't connect to the redis stream?

ruokun-niu: Good points. I'll update this

ruokun-niu: As a starting point, I think we can start with two health checks
1. checks if dapr sidecar is running
2. an azure signalr check that only runs when Azure SignalR is configured (via connection string)

Any thoughts?

reactions/signalr/signalr-reaction/Drasi.Reactions.SignalR/Services/AzureSignalRHealthCheck.cs
```diff
@@ -0,0 +1,127 @@
...
+                return HealthCheckResult.Unhealthy(
+                    "Failed to parse Azure SignalR Service endpoint from connection string");
+            }
+
+            // Try to reach the Azure SignalR Service health endpoint
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(TimeSpan.FromSeconds(10));
+
+            var healthUrl = $"{endpoint}/api/health";
+            var response = await _httpClient.GetAsync(healthUrl, cts.Token);
```
danielgerlag, line 70: Would this catch an auth issue?

danielgerlag (approved):

reactions/signalr/signalr-reaction/Drasi.Reactions.SignalR/Services/AzureSignalRHealthCheck.cs
ruokun-niu: Yes, but I have updated the code to check the status code and treat authentication failures (401/403) as Unhealthy instead of Degraded.

PR #341, opened 2025-11-17
Integrate Drasi Server Core into drasi-platform

agentofreality:
# Description

This PR integrates Drasi Server Core into Drasi Platform and maes it an option for use as a Query Runtime Within the QueryHost for executing Continuous Queries.

### DIFF
--- a/sources/kubernetes/kubernetes-proxy/src/main.rs
+++ b/sources/kubernetes/kubernetes-proxy/src/main.rs
@@ -23,19 +23,19 @@ use tokio::sync::mpsc::UnboundedSender;
 #[tokio::main]
 async fn main() {
     env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
-    
+
     let kube_config = {
         match env::var("kubeConfig") {
             Ok(config) => {
-                let kc = Kubeconfig::from_yaml(&config).unwrap();
+                let kc = Kubeconfig::from_yaml(&config).expect("Failed to parse kubeconfig");
                 Config::from_custom_kubeconfig(kc, &KubeConfigOptions::default())
                     .await
-                    .unwrap()
+                    .expect("Failed to create kube config from kubeconfig")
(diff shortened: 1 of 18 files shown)


query-container/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+rustflags = ["-L", "/opt/homebrew/opt/oniguruma/lib"]
+
+[env]
+JQ_LIB_DIR = "/opt/homebrew/Cellar/jq/1.8.1/lib"
```
danielgerlag, line 5: should probably gitignore this file.

query-container/query-host/examples/query-runtime-config.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+# Query using ServerCore runtime (new, self-contained)
+---
+id: "example-query-servercore"
+spec:
+  mode: "continuous"
+  query: "MATCH (n:Product)-[:SOLD_IN]->(s:Store) RETURN n.name, s.location"
+  queryLanguage: "Cypher"
+  runtime: "ServerCore"  # Use ServerCore runtime
+  sources:
```
danielgerlag, line 35: How do we run native drasi-lib in-process sources & reactions?

query-container/query-host/src/query_actor.rs
```diff
@@ -338,25 +424,245 @@ impl QueryActor {
...
+        self.runtime
+            .set(QueryRuntimeInstance::Worker(Arc::new(worker)))
+            .await;
 
         log::info!("Query {} worker started", self.query_id);
 
         Ok(())
     }
 
+    async fn init_core(&self) -> Result<(), ActorError> {
```
danielgerlag: Would it make sense to have a runtime factory of some kind, that works returns a Runtime trait, which could either be a Worker / ServerCore implementation?

PR #370, opened 2025-12-11
VsCode Extension: Intellisense & Validation for Drasi YAML

danielgerlag:
Adds intellisense and validation for Drasi YAML in Vs Code.

image

image

image


dev-tools/vscode/drasi/README.md
```diff
@@ -25,7 +70,14 @@ The **Drasi** extension provides powerful tools for managing and debugging Drasi
 2. Use the **Workspace Explorer** to browse and manage your YAML files.
 3. Use the **Drasi Explorer** to interact with live resources in your Kubernetes cluster.
 4. Right-click on resources in the explorer or use CodeLens actions to apply, debug, or validate resources.
+5. Start typing in a YAML file - intellisense will automatically activate for Drasi resources.
+
+### Example YAML File
 
+See `example-drasi-resources.yaml` in the extension directory for a complete example demonstrating:
```
ruokun-niu: Was this file checked in?

dev-tools/vscode/drasi/src/yaml-schema-provider.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  private configureYamlExtension() {
+    const config = vscode.workspace.getConfiguration('yaml');
+    
+    // Disable validation (our custom provider handles it)
+    config.update('validate', false, vscode.ConfigurationTarget.Global);
+    
+    // CLEAR ALL existing schema registrations first
+    config.update('schemas', {}, vscode.ConfigurationTarget.Global);
```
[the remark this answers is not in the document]
amansinghoriginal: Will this wipe any existing schema associations the user may have configured for other tools like Kubernetes, Ansible, GitHub actions?

dev-tools/vscode/drasi/src/drasi-completion-provider.ts
```diff
@@ -0,0 +1,186 @@
...
+      // Parse the path from indentation and keys
+      const currentPath = this.parseYamlPath(textBeforeCursor);
+      
+      return { kind, currentPath };
+    } catch (e) {
+      return { kind: null, currentPath: [] };
+    }
+  }
+
+  private parseYamlPath(text: string): string[] {
```
amansinghoriginal: These custom methods rely on Regex and Indentation Counting. Should we instead explore traversing AST created by `yaml` library, which is already imported?

dev-tools/vscode/drasi/schemas/drasi-resources.schema.json
amansinghoriginal, line 1: Can we not register this schema file and use the `redhat.vscode-yaml` directly. Then we would not need to write our own parser or completion provider, right?

ruokun-niu (approved):

dev-tools/vscode/drasi/src/yaml-schema-provider.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Use the union schema that handles all Drasi resource types
+    // This schema uses conditional logic (allOf/if/then) to apply the correct
+    // schema based on the 'kind' field, which works for multi-document files
+    const unionSchemaUri = vscode.Uri.joinPath(this.extensionUri, 'schemas/drasi-resources.schema.json').toString();
+    
+    const existingSchemas = config.get<any>('schemas') || {};
+    
+    const schemas: any = { ...existingSchemas };
+    schemas[unionSchemaUri] = [
+      '**/*query*.yaml',
```
amansinghoriginal, line 42: This will apply Drasi schemas to any file with "query" in the name - is that too broad?

dev-tools/vscode/drasi/schemas/drasi-resources.schema.json
amansinghoriginal, line 1: Should we consider using `$ref` here to reference individual schemas instead of duplicating content?

amansinghoriginal (approved):
Approved with suggestions.

dev-tools/vscode/drasi/src/yaml-schema-provider.ts
danielgerlag: That is looking for the schema itself, rather than targets to apply it to.

PR #384, opened 2026-01-17
fix: return error for unknown resource kind in wait command

### LINKED ISSUE #109: Drasi CLI unresponsive if waiting on invalid resource
### Steps to reproduce

Run the wait command in Drasi CLI with any term other than source/reaction/query like below, and the terminal becomes unresponsive indefinitely.
```sh
drasi wait aman
```

### Observed behavior

_No response_

### Desired behavior

For unknown resource types where Drasi CLI cannot wait, it should stop the wait and respond with the proper error message.

### Workaround

No workaround, have to kill the terminal on MacOS.

### Drasi Version

0.0.14

### Operating system

macOS Sequoa, M1chip

### Additional context

_No response_

- [X] Yes, I would like to support you

shri3016:
# Description

This PR adds validation for unknown resource kinds in the drasi wait command. Previously, running drasi wait with an invalid resource type (e.g., drasi wait aman) would cause the CLI to hang indefinitely because the code would create a malformed URL and wait forever for a response. Now, the CLI validates the resource kind before making the HTTP request and returns a clear error message: Error: unknown resource kind 'aman'. Supported kinds are: source, reaction, query, continuousquery, querycontainer, sourceprovider, reactionprovider

## Type of change

- This pull request fixes a bug in Drasi and has an approved issue (#109 )

Fixes: #109

### DIFF
--- a/cli/sdk/api_client.go
+++ b/cli/sdk/api_client.go
@@ -198,9 +198,18 @@ func (t *ApiClient) ReadyWait(manifests *[]api.Manifest, timeout int32, output o
 	for _, mf := range *manifests {
 		subject := "Wait " + mf.Kind + "/" + mf.Name
 
+		// Validate the resource kind before making the request
+		route, ok := kindRoutes[strings.ToLower(mf.Kind)]
+		if !ok {
+			errMsg := fmt.Sprintf("unknown resource kind '%s'. Supported kinds are: source, reaction, query, continuousquery, querycontainer, sourceprovider, reactionprovider", mf.Kind)
+			output.AddTask(subject, fmt.Sprintf("Waiting for %v/%v to come online", mf.Kind, mf.Name))
+			output.FailTask(subject, fmt.Sprintf("Error: %v", errMsg))
+			return errors.New(errMsg)
+		}
+
 		output.AddTask(subject, fmt.Sprintf("Waiting for %v/%v to come online", mf.Kind, mf.Name))
 
-		url := fmt.Sprintf("%v/%v/%v/%v/ready-wait?timeout=%v", t.prefix, mf.ApiVersion, kindRoutes[strings.ToLower(mf.Kind)], mf.Name, timeout)
+		url := fmt.Sprintf("%v/%v/%v/%v/ready-wait?timeout=%v", t.prefix, mf.ApiVersion, route, mf.Name, timeout)
 
 		req, err := http.NewRequest(http.MethodGet, url, bytes.NewReader([]byte{}))
 		if err != nil {
--- a/cli/cmd/utils.go
+++ b/cli/cmd/utils.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
@@ -69,7 +70,7 @@ func loadManifests(cmd *cobra.Command, args []string) (*[]api.Manifest, error) {
 				Name:       args[1],
 				Spec:       nil,
 			})
-		} else {
+		} else if len(args) == 0 {
 			// Possibility 2: we have no arguments and we need to retrieve the manifests from pipe
 			stat, err := os.Stdin.Stat()
 			if err != nil {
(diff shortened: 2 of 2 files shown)


cli/sdk/api_client.go
amansinghoriginal, line 1: Your changes to cli/sdk/api_client.go are solid.                                                                                                                      
  * The logic to check if mf.Kind exists in kindRoutes is correct and idiomatic Go.
  * The error message you added is very helpful for users because it explicitly lists the supported resource kinds.
  * This improves the robustness of the SDK by preventing invalid requests from being sent to the API.

cli/sdk/api_client.go
```diff
@@ -198,9 +198,18 @@ func (t *ApiClient) ReadyWait(manifests *[]api.Manifest, timeout int32, output o
 	for _, mf := range *manifests {
 		subject := "Wait " + mf.Kind + "/" + mf.Name
 
+		// Validate the resource kind before making the request
+		route, ok := kindRoutes[strings.ToLower(mf.Kind)]
+		if !ok {
+			errMsg := fmt.Sprintf("unknown resource kind '%s'. Supported kinds are: source, reaction, query, continuousquery, querycontainer, sourceprovider, reactionprovider", mf.Kind)
```
amansinghoriginal, line 204:
```suggestion
			errMsg := fmt.Sprintf("Unknown resource kind '%s'. Supported kinds are: source, reaction, query, continuousquery, querycontainer, sourceprovider, reactionprovider", mf.Kind)
```

cli/sdk/api_client.go
amansinghoriginal, line 1: @shri3016 Have you had a chance to run the reproduction step (drasi wait aman) locally with your changes applied?

I suspect you might notice that the CLI still appears unresponsive or exits without printing your new error message.

amansinghoriginal, line 193: For your validation to run, manifests must contain at least one item.

amansinghoriginal, line 1: I suspect that the issue lies upstream in how the CLI parses arguments before it even calls `ReadyWait`. Take a look at the loadManifests function in cli/cmd/utils.go (line 46).

@shri3016 As a challenge - Trace what happens in loadManifests when args has a length of 1 (like in `drasi wait aman`)
  1. Does it enter the if len(args) == 2 block?
  2. Does it return an error, or does it return an empty list of manifests?

amansinghoriginal (changes requested):
Thank you for submitting this pull request. We truly appreciate your contributions to Drasi.

To fully fix the reported bug, we need to ensure that the CLI validation catches this "invalid argument count" scenario before passing an empty list to the SDK.

Keep your current changes! They are a valid improvement for handling "Invalid Kind" scenarios (e.g., `drasi wait invalidKind myName`).

You might need to modify `cli/cmd/utils.go` to return an error if the argument count is incorrect (i.e., not 0 and not 2).

Let me know if you have any questions as you dig into the code. You're doing great! ✨

Also, engage with our community on our [discord server](https://discord.com/invite/AX7FneckBq)

cli/sdk/api_client.go
shri3016: Thank you!

shri3016: I've now added argument count validation in cli/cmd/utils.go that returns an error when args is not 0 or 2. After this fix, drasi wait aman correctly shows: Error: invalid argument count: expected 0 or 2 arguments, got 1,Attaching the screenshot below.
error message

shri3016: I have added validation in cli/cmd/utils.go to catch invalid argument counts before an empty list reaches ReadyWait. Now the error is returned early in loadManifests.

shri3016: Traced it! When len(args) == 1: - It does not enter the if len(args) == 2 block - It was returning an empty list (no error). Fixed by adding an else block that returns an error when args count is not 0 or 2.
Thanks for the guidance @amansinghoriginal . Please let me know if I have to make any changes whenever you have time to review this.

amansinghoriginal (approved):
Thank you for your patience and for your first contribution to Drasi!

PR #388, opened 2026-01-30
feat: Add Oracle database support for relational sources

### LINKED ISSUE #154: add oracle source Support
### Overview of feature request

Code refactoring in [PR #146](https://github.com/drasi-project/drasi-platform/pull/146) has added MySQL support. Would it be easy to add Oracle support based on this? Are there plans to support more databases?

### Acceptance criteria

_No response_

### Additional context

_No response_

- [ ] Yes, I would like to support you

Eeshu-Yadav:
# Description
Add Oracle database support for relational sources

- Add Oracle Debezium connector dependency to debezium-reactivator
- Add Oracle JDBC driver dependency to sql-proxy
- Implement Oracle database strategy with LogMiner CDC support
- Add Oracle connection handling with SID/service name support
- Update DebeziumReactivator to handle Oracle connector
- Add Oracle source provider configuration to CLI
- Create end-to-end test scenario for Oracle integration

Fixes: #154

### DIFF
--- a/cli/installers/resources/default-source-providers.yaml
+++ b/cli/installers/resources/default-source-providers.yaml
@@ -169,6 +169,57 @@ spec:
 ---
 apiVersion: v1
 kind: SourceProvider
+name: Oracle
+spec:
+  services:
+    proxy:
+      image: source-sql-proxy
+      dapr:
+        app-port: "4002"
+      config_schema:
+        type: object
+        properties:
+          connector:
+            type: string
+            default: Oracle
+    reactivator:
+      image: source-debezium-reactivator
+      deprovisionHandler: true
+      dapr:
+        app-port: "80"
+      config_schema:
+        type: object
+        properties:
+          connector:
+            type: string
+            default: Oracle
+  config_schema:
+    type: object
+    properties:
+      database:
+        type: string
+      serviceName:
+        type: string
+      host:
+        type: string
+      password:
+        type: string
+      port:
+        type: number
+        default: 1521
+      user:
+        type: string
+      tables:
+        type: array
+    required:
+      - host
+      - port
+      - user
+      - password
+      - tables
+---
+apiVersion: v1
+kind: SourceProvider
 name: CosmosGremlin
 spec:
   config_schema:
--- a/cli/installers/resources/default-source-providers.yaml
+++ b/cli/installers/resources/default-source-providers.yaml
@@ -198,8 +198,10 @@ spec:
     properties:
       database:
         type: string
+        description: The CDB name (e.g. XE). Required for all Oracle setups.
       serviceName:
         type: string
+        description: The PDB name for CDB/PDB (multi-tenant) setups (e.g. XEPDB1). Optional for non-CDB installations.
       host:
         type: string
       password:
@@ -212,6 +214,7 @@ spec:
       tables:
         type: array
     required:
+      - database
       - host
       - port
       - user
--- a/e2e-tests/10-oracle-scenario/oracle.test.js
+++ b/e2e-tests/10-oracle-scenario/oracle.test.js
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2024 The Drasi Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { afterAll, beforeAll, describe, expect, test } from '@jest/globals';
+import { deployFixture } from '../fixtures/deploy-resources';
+import { undeployFixture } from '../fixtures/delete-resources';
+import { SignalRFixture } from '../fixtures/signalr-fixture';
+
+describe('Oracle Database Integration', () => {
+  const scenarioName = '10-oracle-scenario';
+  let signalRFixture;
+
+  beforeAll(async () => {
+    await deployFixture(scenarioName);
+    signalRFixture = new SignalRFixture(scenarioName);
+    await signalRFixture.start();
+  }, 120000);
+
+  afterAll(async () => {
+    if (signalRFixture) {
+      await signalRFixture.stop();
+    }
+    await undeployFixture(scenarioName);
+  });
+
+  test('Oracle source should connect and retrieve data', async () => {
+    const queryId = 'oracle-query1';
+
+    const result = await signalRFixture.waitForQueryResult(queryId);
+    expect(result).toBeDefined();
+    expect(result.length).toBeGreaterThan(0);
+
+    const item = result[0];
+    expect(item).toHaveProperty('Id');
+    expect(item).toHaveProperty('Name');
+    expect(item).toHaveProperty('Category');
+    expect(item).toHaveProperty('Score');
+
+    expect(item.Category).toBe('1');
+    expect(item.Score).toBeGreaterThanOrEqual(5);
+  }, 30000);
+
+  test('Oracle change data capture should work', async () => {
+    // This test would need to connect to Oracle and make changes
+    // For now, we'll just verify the query is active and returns data
+    const queryId = 'oracle-query1';
+    const result = await signalRFixture.waitForQueryResult(queryId);
+    expect(result).toBeDefined();
+    expect(result.length).toBeGreaterThan(0);
+  }, 30000);
+});
--- a/e2e-tests/10-oracle-scenario/oracle.test.js
+++ b/e2e-tests/10-oracle-scenario/oracle.test.js
@@ -14,51 +14,63 @@
  * limitations under the License.
  */
 
-import { afterAll, beforeAll, describe, expect, test } from '@jest/globals';
-import { deployFixture } from '../fixtures/deploy-resources';
-import { undeployFixture } from '../fixtures/delete-resources';
-import { SignalRFixture } from '../fixtures/signalr-fixture';
+const yaml = require('js-yaml');
+const fs = require('fs');
+const deployResources = require('../fixtures/deploy-resources');
+const deleteResources = require('../fixtures/delete-resources');
+const PortForward = require('../fixtures/port-forward');
+const SignalRFixture = require('../fixtures/signalr-fixture');
+const oracledb = require('oracledb');
 
-describe('Oracle Database Integration', () => {
-  const scenarioName = '10-oracle-scenario';
-  let signalRFixture;
+let dbPortForward = new PortForward('oracle', 1521);
+let signalrFixture = new SignalRFixture(['oracle-query1']);
+let dbConn;
 
-  beforeAll(async () => {
-    await deployFixture(scenarioName);
-    signalRFixture = new SignalRFixture(scenarioName);
-    await signalRFixture.start();
-  }, 120000);
+beforeAll(async () => {
+  const resources = yaml.loadAll(fs.readFileSync(__dirname + '/resources.yaml', 'utf8'));
+  await deployResources(resources);
+  await signalrFixture.start();
 
-  afterAll(async () => {
-    if (signalRFixture) {
-      await signalRFixture.stop();
-    }
-    await undeployFixture(scenarioName);
+  const localPort = await dbPortForward.start();
+  dbConn = await oracledb.getConnection({
+    user: 'testuser',
+    password: 'oracle',
+    connectString: `127.0.0.1:${localPort}/XEPDB1`,
   });
 
-  test('Oracle source should connect and retrieve data', async () => {
-    const queryId = 'oracle-query1';
+  await new Promise(r => setTimeout(r, 15000)); // reactivator is slow to startup
+}, 180000);
 
-    const result = await signalRFixture.waitForQueryResult(queryId);
-    expect(result).toBeDefined();
-    expect(result.length).toBeGreaterThan(0);
+afterAll(async () => {
+  await signalrFixture.stop();
+  if (dbConn) {
+    await dbConn.close();
+  }
+  dbPortForward.stop();
 
-    const item = result[0];
-    expect(item).toHaveProperty('Id');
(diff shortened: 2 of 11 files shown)


e2e-tests/10-oracle-scenario/oracle.test.js
```diff
@@ -0,0 +1,59 @@
...
+    
+    expect(item.Category).toBe('1');
+    expect(item.Score).toBeGreaterThanOrEqual(5);
+  }, 30000);
+
+  test('Oracle change data capture should work', async () => {
+    // This test would need to connect to Oracle and make changes
+    // For now, we'll just verify the query is active
+    const queryId = 'oracle-query1';
+    expect(queryResults[queryId]).toBeDefined();
```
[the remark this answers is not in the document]
danielgerlag: agreed

e2e-tests/10-oracle-scenario/oracle.test.js
```diff
@@ -0,0 +1,59 @@
...
+  }, 15000);
```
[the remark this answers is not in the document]
danielgerlag: agreed

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
```diff
@@ -0,0 +1,137 @@
...
+        try (var stmt = conn.prepareStatement("ALTER DATABASE ADD SUPPLEMENTAL LOG DATA")) {
+            stmt.execute();
+            log.info("Enabled minimal supplemental logging for database");
+        } catch (SQLException e) {
+            log.warn("Could not enable supplemental logging: {}", e.getMessage());
+        }
+    }
+
+    private void enableTableSupplementalLogging(Connection conn, String tableName) throws SQLException {
+        try (var stmt = conn.prepareStatement("ALTER TABLE " + tableName + " ADD SUPPLEMENTAL LOG DATA (ALL) COLUMNS")) {
```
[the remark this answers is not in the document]
danielgerlag: agreed

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
```diff
@@ -0,0 +1,137 @@
...
+            mapping.keyField = rs.getString("COLUMN_NAME");
+            mapping.labels = Collections.singleton(tableName);
+            
+            return mapping;
+        }
+    }
+
+    @Override
+    public void initialize(Configuration config, RelationalGraphMapping relationalGraphMapping) {
+        try (var conn = getConnection(config).connection()) {
```
danielgerlag: Resource leak in Oracle.java:initialize()

```java
   try (var conn = getConnection(config).connection()) { ... }
```
  The intermediate JdbcConnection (returned by getConnection()) is never closed. Only the inner java.sql.Connection is auto-closed.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public Configuration createConnectorConfig(Configuration baseConfig) {
+        return Configuration.create()
+                .with(baseConfig)
+                .with("connector.class", "io.debezium.connector.oracle.OracleConnector")
+                .with("log.mining.strategy", "online_catalog")
+                .with("snapshot.mode", "no_data")
+                .with("log.mining.continuous.mine", "true")
+                .with("log.mining.batch.size.default", "1000")
```
danielgerlag: The Oracle Debezium connector requires explicitly setting database.connection.adapter to "logminer" (or "xstream" / "olr").

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public Configuration createConnectorConfig(Configuration baseConfig) {
+        return Configuration.create()
+                .with(baseConfig)
+                .with("connector.class", "io.debezium.connector.oracle.OracleConnector")
+                .with("log.mining.strategy", "online_catalog")
+                .with("snapshot.mode", "no_data")
+                .with("log.mining.continuous.mine", "true")
```
danielgerlag: log.mining.continuous.mine is deprecated

sources/relational/ORACLE_SUPPORT.md
```diff
@@ -0,0 +1,117 @@
+# Oracle Database Source Support
```
danielgerlag: Let's remove the AI summary, and create a PR in the docs repo

ORACLE_IMPLEMENTATION_SUMMARY.md
```diff
@@ -0,0 +1,188 @@
+# Oracle Database Support Implementation Summary
```
danielgerlag: Let's remove the AI summary

danielgerlag (changes requested):

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
```diff
@@ -0,0 +1,156 @@
...
+        return Configuration.create()
+                .with(baseConfig)
+                .with("connector.class", "io.debezium.connector.oracle.OracleConnector")
+                .with("log.mining.strategy", "online_catalog")
+                .with("snapshot.mode", "no_data")
+                .with("database.connection.adapter", "logminer")
+                .with("log.mining.batch.size.default", "1000")
+            .build();
+    }
+
```
amansinghoriginal: For CDB/PDB (multi-tenant) Oracle setups, Debezium needs `database.pdb.name` to know which Pluggable Database to monitor via LogMiner. Right now, the `serviceName` property from the SourceProvider config is only used in the SQL Proxy's JDBC URL but never reaches the reactivator.

Could you wire it through here when it's provided? Something like:

```java
var builder = Configuration.create()
        .with(baseConfig)
        .with("connector.class", "io.debezium.connector.oracle.OracleConnector")
        .with("log.mining.strategy", "online_catalog")
        .with("snapshot.mode", "no_data")
        .with("database.connection.adapter", "logminer")
        .with("log.mining.batch.size.default", "1000");

var pdbName = Reactivator.GetConfigValue("serviceName", "");
if (!pdbName.isEmpty()) {
    builder = builder.with("database.pdb.name", pdbName);
}

return builder.build();
```

This should stay optional since non-CDB Oracle installations don't need it. But note that Oracle 12c+ defaults to CDB/PDB architecture, and your e2e test uses Oracle XE 21c with PDB `XEPDB1`, so it would be needed there.

When using CDB/PDB, the `database` property should be the CDB name (e.g., `XE`) and `serviceName` the PDB name (e.g., `XEPDB1`). It might be worth noting that in the SourceProvider schema description as well.

amansinghoriginal, line 94: For CDB/PDB (multi-tenant) Oracle setups, Debezium needs `database.pdb.name` to know which Pluggable Database to monitor via LogMiner. Right now, the `serviceName` property from the SourceProvider config is only used in the SQL Proxy's JDBC URL but never reaches the reactivator.

Could you wire it through here when it's provided? Something like:

```java
var builder = Configuration.create()
        .with(baseConfig)
        .with("connector.class", "io.debezium.connector.oracle.OracleConnector")
        .with("log.mining.strategy", "online_catalog")
        .with("snapshot.mode", "no_data")
        .with("database.connection.adapter", "logminer")
        .with("log.mining.batch.size.default", "1000");

var pdbName = Reactivator.GetConfigValue("serviceName", "");
if (!pdbName.isEmpty()) {
    builder = builder.with("database.pdb.name", pdbName);
}

return builder.build();
```

This should stay optional since non-CDB Oracle installations don't need it. But note that Oracle 12c+ defaults to CDB/PDB architecture, and your e2e test uses Oracle XE 21c with PDB `XEPDB1`, so it would be needed there.

When using CDB/PDB, the `database` property should be the CDB name (e.g., `XE`) and `serviceName` the PDB name (e.g., `XEPDB1`). It might be worth noting that in the SourceProvider schema description as well.

cli/installers/resources/default-source-providers.yaml
```diff
@@ -169,6 +169,57 @@ spec:
...
+      user:
+        type: string
+      tables:
+        type: array
+    required:
+      - host
+      - port
+      - user
+      - password
+      - tables
```
amansinghoriginal, line 222: `database` should be added to the `required` list here. The reactivator's `RelationalChangeMonitor.createBaseConfig()` (line 94-96) calls `Reactivator.TerminalError` if `database` is null or empty, so this is a hard requirement at runtime.

All three other relational providers (PostgreSQL, MySQL, SQLServer) require `database` in their schemas. Without it, a user can submit a Source manifest that passes schema validation but crashes the reactivator at startup.

The e2e test in this PR also doesn't set `database`, so it would hit this error.

```yaml
    required:
      - database
      - host
      - port
      - user
      - password
      - tables
```

e2e-tests/10-oracle-scenario/oracle.test.js
```diff
@@ -0,0 +1,64 @@
...
+    expect(result).toBeDefined();
+    expect(result.length).toBeGreaterThan(0);
+
+    const item = result[0];
+    expect(item).toHaveProperty('Id');
+    expect(item).toHaveProperty('Name');
+    expect(item).toHaveProperty('Category');
+    expect(item).toHaveProperty('Score');
+
+    expect(item.Category).toBe('1');
```
amansinghoriginal: Thanks for getting the test scaffolding in place! We do need a bit more coverage here before we can merge though.

Right now both tests only verify bootstrap data (initial snapshot). For us to have confidence that the Oracle CDC pipeline actually works end-to-end, we need at least one test that performs a database mutation and verifies the change propagates. The existing PostgreSQL test in `01-simple-scenario/my.test.js` is a good reference for the pattern:

1. Connect to the database directly (you would need to add `oracledb` as a dependency in `e2e-tests/package.json`)
2. Perform an `UPDATE` or `INSERT`
3. Call `signalRFixture.waitForChange()` and assert the change event has the correct `op` and values

This is how we verify that the full pipeline (Debezium Reactivator -> Change Router -> Change Dispatcher -> Query Container) is working, not just the SQL Proxy bootstrap path.

Happy to help if you run into any issues getting `oracledb` set up for the test!

amansinghoriginal (changes requested):
Hey @Eeshu-Yadav, thank you so much for picking up this issue and for your patience while we got to the review. We really appreciate every contribution to Drasi and this is a great addition to our source ecosystem!                                                                                                    
                                                                  
Your implementation follows the existing strategy pattern nicely and you've clearly done your homework on how the relational sources are structured. Nice work on getting the Debezium LogMiner config, the SQL Proxy connection, and the SourceProvider definition all wired up.


I've left three comments on areas that need adjustment before we can merge:

  1. PDB support in the reactivator - the serviceName config needs to be passed through to Debezium as database.pdb.name for CDB/PDB setups (the standard Oracle12c+ architecture)

  2. database must be required in the SourceProvider schema - the reactivator will crash at startup without it

  3. E2e test coverage - we need at least one test that actually mutates the database and verifies the change event propagates, following the pattern in 01-simple-scenario

I also noticed that Daniel's earlier feedback on the CDC test and a few other items is still outstanding. Please take a look at those comments as well.

The bones of the PR are solid and you're close! Let us know if you have any questions or need a hand with any of the changes.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
Eeshu-Yadav: Good catch! Fixed — when `serviceName` is present in the config, it now gets passed as `database.pdb.name` to the Debezium builder. Left it optional so non-CDB setups still work. Also updated `resources.yaml` to use `database: XE` as the CDB name and `serviceName: XEPDB1` as the PDB, and added a note in the schema descriptions to clarify the distinction.

cli/installers/resources/default-source-providers.yaml
Eeshu-Yadav: Done, added `database` to the required list. Makes sense — without it the reactivator just crashes at startup which is a really bad experience. I also added short descriptions to `database` and `serviceName` in the schema so it's clearer which is the CDB name and which is the PDB name.

e2e-tests/10-oracle-scenario/oracle.test.js
Eeshu-Yadav: Thanks for the detailed feedback and the pointer to `01-simple-scenario`! Rewrote the test completely — it now uses `deployResources`/`deleteResources` properly, sets up a port-forward to Oracle, connects with `oracledb`, and has two tests: one for the bootstrap snapshot via `requestReload`, and a proper CDC test that does an UPDATE + commit and waits for the change event via `waitForChange`. Also added `oracledb` as a dependency in `package.json`.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
Eeshu-Yadav: Good catch, thanks! Fixed — `initialize()` now declares both the `JdbcConnection` and the inner `java.sql.Connection` in the try-with-resources block so both get properly closed when the block exits.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
Eeshu-Yadav: Added — `database.connection.adapter` is now explicitly set to `logminer` in `createConnectorConfig`.

sources/relational/debezium-reactivator/src/main/java/io/drasi/databases/Oracle.java
Eeshu-Yadav: Removed that property, thanks for flagging it.

sources/relational/ORACLE_SUPPORT.md
Eeshu-Yadav: Removed from the PR. Will follow up with a separate PR to the docs repo to add proper Oracle source documentation there.

ORACLE_IMPLEMENTATION_SUMMARY.md
Eeshu-Yadav: Removed, thanks for the feedback.

PR #389, opened 2026-02-01
Fix Redis connection panic loop in query worker

remo-lab:
### Summary

This PR fixes a panic/restart loop in the query worker that happens when Redis is temporarily unavailable. The issue was caused by an `.unwrap()` during worker initialization, which caused the worker to panic instead of failing and retrying cleanly.

---

## What was going wrong

While looking into Redis outage behavior, I noticed that if Redis is briefly unreachable (network blip, pod restart, etc.), the query worker crashes in a loop:

- Worker initialization panics on a failed Redis connection
- The process stays alive, but the worker task dies
- The actor reminder fires and tries to re-initialize the worker
- The same `.unwrap()` is hit again and panics again

This continues until Redis comes back up. During this time, queries in that container are effectively unavailable and logs get spammed with panics.

---

## Root cause

In `query_worker.rs`, the Redis publisher connection was using an unsafe unwrap:

```rust
Publisher::connect(...).await.unwrap(); // todo
```

So any transient Redis failure immediately caused a panic. The `// todo` comment suggests this was already known but never handled.

---

## Fix

I replaced the unwrap with explicit error handling. On failure, the worker now:

- Logs the Redis connection error
- Transitions to `QueryState::TransientError`
- Exits initialization cleanly

This allows the existing reminder-based retry logic to retry without crashing the worker.

No new retry logic was added — this just lets the existing mechanism work as intended.

---

## Changes

- File: `query-container/query-host/src/query_worker.rs`
- Replaced `.unwrap()` with proper error handling
- Small, contained change (no API changes, no new deps)

---

## Testing / Impact

- Existing retry path was already working; this enables it for Redis failures
- Prevents panic loops during Redis outages
- Workers recover automatically once Redis is available again

---

## Checklist

- [x] DCO sign-off included  
- [x] Follows existing error handling patterns  
- [x] No breaking changes  
- [x] Single, focused fix

### DIFF
--- a/query-container/query-host/src/query_worker.rs
+++ b/query-container/query-host/src/query_worker.rs
@@ -179,12 +179,19 @@ impl QueryWorker {
 
             fill_default_source_labels(&mut modified_config, &continuous_query.get_query());
 
-            let source_publisher = change_stream::publisher::Publisher::connect(
+            let source_publisher = match change_stream::publisher::Publisher::connect(
                 &stream_config.redis_url,
                 topic.clone(),
             )
             .await
-            .unwrap(); //todo
+            {
+                Ok(p) => p,
+                Err(err) => {
+                    log::error!("Failed to connect to Redis publisher: {}", err);
+                    lifecycle.change_state(QueryState::TransientError(err.to_string()));
+                    return;
+                }
+            };
 
             let future_consumer =
                 FutureConsumer::new(Arc::new(source_publisher), query_id.to_string());
--- a/query-container/query-host/src/query_worker.rs
+++ b/query-container/query-host/src/query_worker.rs
@@ -187,7 +187,7 @@ impl QueryWorker {
             {
                 Ok(p) => p,
                 Err(err) => {
-                    log::error!("Failed to connect to Redis publisher: {}", err);
+                    log::error!("Failed to connect to source publisher: {}", err);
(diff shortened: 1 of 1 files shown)


remo-lab:
HI @danielgerlag @agentofreality ,This fixes a panic loop in the query worker when Redis is temporarily unavailable by replacing an unsafe unwrap with proper error handling. It allows the existing retry mechanism to work as intended.

Would really appreciate a review when you have some time. Thanks!

query-container/query-host/src/query_worker.rs
```diff
@@ -179,12 +179,19 @@ impl QueryWorker {
...
+            let source_publisher = match change_stream::publisher::Publisher::connect(
                 &stream_config.redis_url,
                 topic.clone(),
             )
             .await
-            .unwrap(); //todo
+            {
+                Ok(p) => p,
+                Err(err) => {
+                    log::error!("Failed to connect to Redis publisher: {}", err);
```
danielgerlag: The publisher component could be swapped out for a different implementation, eg. kafka, rabbitmq, etc.

```suggestion
                    log::error!("Failed to connect to source publisher: {}", err);
```

danielgerlag (approved):

ruokun-niu (approved):

Issue #411, opened 2026-03-14
Reaction SDK(.NET): Add explicit Unit Tests for ManagementClient and ResultViewClient


amansinghoriginal:
### ⚠️ Contributor & GSoC Rules ⚠️
To ensure a fair and manageable process for all contributors (including Google Summer of Code applicants), please follow these rules:
1. **Do NOT open a Pull Request unless you have been assigned to this issue.** Unassigned PRs will be closed without review.
2. **To get assigned:** Leave a comment below explaining *where* in the codebase you will implement this, and a 1-2 sentence description of your approach. The maintainers will assign it to the first coherent proposal.
3. You may only be assigned to one "good first issue" at a time.
4. If there is no update within 4 days, we may have to reassign the issue to another contributor.

---

### Area for Improvement
Currently, the `.NET Reaction SDK` has a test project located at `reactions/sdk/dotnet/Drasi.Reaction.SDK.Tests/`. However, it only contains tests for the core `ReactionTests` (like the `ReactionBuilder`).

The core services located in `reactions/sdk/dotnet/Drasi.Reaction.SDK/Services/`, specifically the `ManagementClient` and `ResultViewClient`, completely lack unit test coverage. This leaves a gap where networking logic, API endpoint paths, and JSON deserialization might silently break during future refactoring.

### Desired behavior
The test suite in `Drasi.Reaction.SDK.Tests/` is expanded to include comprehensive unit tests for both `ManagementClient` and `ResultViewClient`, ensuring our HTTP integrations with the Drasi backend are robust and tested.

hungnguyen1607:
Hi @amansinghoriginal , I am planning to make 2 new test files ManagementClientTests.cs and ResultViewClientTests.cs that utilize custom MockHttpMessageHandler to basically simulate the HTTP responses, essentially helping to do the unit test on the API endpoint paths, json deserialization etc.

Please let me know If i can take care of this issue. Thank you!!!

Shanzita:
Hello @amansinghoriginal sir, I am currently waiting to be assigned on #413 (Python ResultViewClient) and, while digging deep into the SDK for that, this issue also caught my attention.
One thing I noticed is that both ManagementClient and ResultViewClient are currently creating their own HttpClient internally, and this makes them a bit tricky to unit test without some changes to the class first. I'd like to change this to inject HttpClient via the constructor, and this is the standard approach in .NET.
[...]

PR #422, opened 2026-04-12
unit tests for ManagementClient and ResultViewClient #411

hungnguyen1607:
Refactor ManagementClient and ResultViewClient to accept HttpClient via dependency injection, enabling the unit test coverage. 

Both services were creating HttpClient internally earlier, making them untestable.

### DIFF
--- a/reactions/sdk/dotnet/Drasi.Reaction.SDK/Services/ResultViewClient.cs
+++ b/reactions/sdk/dotnet/Drasi.Reaction.SDK/Services/ResultViewClient.cs
@@ -24,7 +24,7 @@ public class ResultViewClient : IResultViewClient
     private readonly IManagementClient _managementClient;
 
     public ResultViewClient(IManagementClient managementClient)
-        : this(new HttpClient(), managementClient)
+    : this(new HttpClient(), managementClient)
     {
     }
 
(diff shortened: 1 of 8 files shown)


reactions/sdk/javascript/package-lock.json
ruokun-niu, line 1: Why was this updated?

hungnguyen1607: Hello, I am back again. 

Sorry for this late update. I have been crushed by my exams for the past weeks.. But I was just playing around with the code and forgot to return it to the original, I remove the changes, please check again.

ruokun-niu (approved):

reactions/sdk/dotnet/Drasi.Reaction.SDK/Services/ResultViewClient.cs
```diff
@@ -23,10 +23,10 @@ public class ResultViewClient : IResultViewClient
     private readonly HttpClient _httpClient;
     private readonly IManagementClient _managementClient;
 
-    public ResultViewClient(IManagementClient managementClient)
+    public ResultViewClient(HttpClient httpClient, IManagementClient managementClient)
```
danielgerlag, line 31: Is this a breaking change?

hungnguyen1607: Yes to the existing code that is passing 1 argument at the time.

hungnguyen1607: Hello!

I  took a closer look and noticed the problem as mentioned, which is passing 1 parameter at the time only! and I replaced the that constructor with a 2-parameter one, which will cause a breaking change as any existing code calling this new ResultViewClient(managementClient) would no longer properly compile.

I added an overload constructor to fix this! (Line 27-29). Essentially keeping the original 1-parameter constructor so existing code can still run, and it will just call the new constructor with a default HttpClient internally

PR #424, opened 2026-04-14
feat(cli): add pre-flight manifest validation to drasi apply

### LINKED ISSUE #423: cli: Add preflight YAML validation in `drasi apply` driven by the management API's OpenAPI spec
### Overview of feature request

Currently, when a user runs `drasi apply` with an invalid or malformed YAML file, the CLI passes it directly to `client.Apply()` with no pre-checks. If the YAML is missing required fields or uses an unrecognized `kind`, the raw HTTP error from the management API bubbles up with no actionable context:

```
Apply: Source/: 404 Not Found
```

The resource name appears empty because the YAML wasn't recognized, and the raw HTTP status gives the user no guidance on what to fix.

**What we are proposing:**
Add a preflight validation step in `cli/cmd/apply.go` **before** `client.Apply()` is called, using `openapi.yaml` — the auto-generated OpenAPI spec for the management API — as the single source of truth for valid resource schemas.

This approach ensures the CLI validation never diverges from what the management API actually accepts, even as new resource kinds or required fields are added in future releases. The `openapi.yaml` is regenerated from the Rust source on every commit via a pre-commit hook, so it is always up to date.

The validator should check each resource in the manifest against its corresponding DTO schema:

| YAML `kind` | API endpoint | Spec schema |
|---|---|---|
| `Source` | `PUT /v1/sources/{id}` | `SourceSpecDto` |
| `Reaction` | `PUT /v1/reactions/{id}` | `ReactionSpecDto` |
| `ContinuousQuery` | `PUT /v1/continuousQueries/{id}` | `QuerySpecDto` |
| `QueryContainer` | `PUT /v1/queryContainers/{id}` | `QueryContainerSpecDto` |
| `SourceProvider` | `PUT /v1/sourceProviders/{id}` | `ProviderSpecDto` |
| `ReactionProvider` | `PUT /v1/reactionProviders/{id}` | `ProviderSpecDto` |

Specifically, the validator should check:
1. `kind` is present and is one of the six recognized resource types
2. `name` is present and non-empty — this becomes the `{id}` path parameter, and an empty name is the direct cause of the `Source/: 404 Not Found` error
3. Required `spec` fields are present per the relevant DTO schema in `openapi.yaml`:
   - `Source`: `kind`
   - `Reaction`: `kind`, `queries`
   - `ContinuousQuery`: `mode`, `query`, `sources` (with `sources.subscriptions`)
   - `QueryContainer`: `queryHostCount`, `results`, `storage`, `defaultStore`
   - `SourceProvider` / `ReactionProvider`: `services`

**Error output — before vs after:**

Before:
```
Apply: Source/: 404 Not Found
```

After:
```
Error: invalid resource at index 0 (kind: Source)
  - missing required field: 'name'

Error: invalid resource at index 0
  - 'kind' must be one of: Source, Reaction, ContinuousQuery, QueryContainer, SourceProvider, ReactionProvider

Error: invalid resource at index 0 (kind: Reaction, name: my-reaction)
  - spec: missing required field 'queries'
```

### Acceptance criteria

- [ ] `drasi apply` performs preflight validation before making any API call
- [ ] Validation is driven by `openapi.yaml` — no hardcoded schema knowledge in the validator
- [ ] Clear, actionable error messages for:
  - Missing `kind`
  - Unrecognized `kind`
  - Missing or empty `name`
  - Missing required `spec` fields per resource type
- [ ] Valid YAML manifests continue to work exactly as before — no regression
- [ ] Unit tests cover valid and invalid cases for each of the six resource types

### Additional context

This was originally reported when a user ran `drasi apply` with a non-Drasi YAML and received a confusing `Apply: Source/: 404 Not Found` error. The root cause is that the resource `name` (used as the `{id}` path parameter in the API call) is empty when the YAML is not a valid Drasi resource, resulting in a path like `Source/` with no ID.

The fix was discussed in the issue thread and the approach of using `openapi.yaml` as the validation source of truth was suggested and approved by @ruokun-niu:
> _"I'd suggest taking a look at our openapi.yaml file, which is the auto-generated OpenAPI spec for the management API. It defines the schemas for all resource types (SourceSpecDto, ReactionSpecDto, QuerySpecDto, etc.) including required fields and valid enum values. This spec is regenerated from the Rust source code on every commit via a pre-commit hook, so it's always up to date."_

**Files involved:**
- `cli/cmd/apply.go` — add `validateManifests(manifests)` call before `client.Apply()`
- `cli/pkg/validator/` (new package) — validation logic reading from `openapi.yaml`
- `openapi.yaml` — management API spec, auto-regenerated from Rust source on every commit

**Related discussion:** #376 

- [ ] Yes, I would like to support you

nXtCyberNet:
# Description

Adds a `validator` package that validates manifests against the embedded `openapi.yaml` 
schema before `client.Apply()` is called. Invalid manifests now print actionable errors 
and abort early — no API call is made.

Previously, `drasi apply` sent manifests directly to the API with no client-side 
validation, producing cryptic errors for missing fields and silently passing everything 
through if `openapi.yaml` was unparseable.

### Changes
- **`cmd/validator.go`** — new package with schema-driven required field checking, 
  two-tier severity (`SeverityWarning` / `SeverityError`), `HasErrors()` helper, and 
  per-kind extensible validators. Never writes to stderr directly — all issues returned 
  as structured `ValidationError`.
- **`cmd/apply.go`** — validation now runs between manifest parsing and `client.Apply()`. 
  Passes `[]api.Manifest` directly . Uses `cmd.PrintErrf()` 
  for capturable output.
- **`cmd/validator_test.go`** — 25 test cases covering valid manifests, missing 
  fields, wrong types, zero-value fields, unknown kinds, bad spec bytes, and error 
  formatting. Uses inline `openapi.yaml` fixture — no file I/O.

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other 
  maintenance task and doesn't change the functionality of Drasi (issue link optional).

Fixes: #423

### DIFF
--- a/cli/cmd/apply.go
+++ b/cli/cmd/apply.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 
 	"drasi.io/cli/api"
 	"drasi.io/cli/output"
@@ -27,6 +28,7 @@ import (
 type applyCmdOptions struct {
 	platformClientFactory func(namespace string) (sdk.PlatformClient, error)
 	outputFactory         func() output.TaskOutput
+	openapiSpec           []byte
 }
 
 // NewApplyCommand creates the apply command with optional dependency injection
@@ -62,6 +64,19 @@ Usage examples:
 
 			if len(*manifests) == 0 {
 				return errors.New("no manifests found. Did you forget to specify the '-f' flag")
+			}
+			validationErrs, err := ValidateManifests(*manifests, opt.openapiSpec)
+
+			if err != nil {
+				return fmt.Errorf("validation setup failed: %w", err)
+			}
+
+			for _, e := range validationErrs {
+				cmd.PrintErrf("%s\n", e.Error())
+			}
+
+			if HasErrors(validationErrs) {
+				return errors.New("apply aborted: one or more manifests failed validation (see above)")
 			}
 
 			var namespace string
--- a/cli/cmd/apply_test.go
+++ b/cli/cmd/apply_test.go
@@ -15,17 +15,17 @@
 package cmd
 
 import (
+	"errors"
+	"os"
+	"testing"
+
 	"drasi.io/cli/api"
 	"drasi.io/cli/output"
 	"drasi.io/cli/sdk"
 	"drasi.io/cli/testutil"
-	"errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
-	"os"
-	"testing"
 )
-
 
 func TestApplyCommand(t *testing.T) {
 	t.Run("Success case - apply single manifest", func(t *testing.T) {
@@ -242,7 +242,8 @@ spec:
 
 		manifestContent := `apiVersion: v1
 kind: Source
-name: test-source`
+name: test-source
+spec: {}`
 		_, err = tempFile.WriteString(manifestContent)
 		assert.NoError(t, err)
 		tempFile.Close()
@@ -275,7 +276,8 @@ name: test-source`
 
 		manifestContent := `apiVersion: v1
 kind: Source
-name: test-source`
+name: test-source
+spec: {}`
 		_, err = tempFile.WriteString(manifestContent)
 		assert.NoError(t, err)
 		tempFile.Close()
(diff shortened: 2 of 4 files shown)


nXtCyberNet:
hi any updates?

PR #437, opened 2026-06-19
feat(python-sdk): export unpacked Pydantic models

### LINKED ISSUE #436: feat(python-sdk): export unpacked Pydantic models
The Dapr Agents extension currently consumes unpacked events published by the Drasi `PostDaprPubSub` reaction. Ideally, the Python reaction SDK should export unpacked models for downstream services to use so they don't need to manually define these models (which are autogenerated).

JeffreyJPZ:
# Description
- Update `generate_types` Makefile target to autogenerate unpacked models
- Bump FastAPI version to `>=0.116.0` to avoid dependency conflict with Dapr Agents

## Type of change

-  [x] This pull request adds or changes features of Drasi and has an approved issue (issue link required).

Closes: #436


amansinghoriginal (changes requested):
1. Did you run make generate-types and then actually import/validate the models?
    I think the committed state won't import.

2. Can we add a trivial `model_validate` test over i/u/d/x JSON?

3.  the regenerated models use module-style relative imports (from . import ChangeSource then use ChangeSource as a type) against an empty models/__init__.py. So the annotations resolve to modules, not classes, and no Pydantic model builds on this PR branch right now. so this would break every existing Python reaction, not only the new unpacked models

@JeffreyJPZ do you think it is being caused by generating three schemas into one flat directory via three sequential datamodel-codegen runs? How does the .NET SDK do this?

JeffreyJPZ (commented):
Yeah, this is definitely something that needs to be fixed since we can't do `from drasi.reaction.models.unpacked import ChangeNotification` as nothing is being exported, will look into it and update here.

Update: looks like there's a flag for `datamodel-code-generator` called `--all-exports-scope` to automatically populate `__all__`, but it was introduced in **0.39.0** so we'll have to bump it to that at the very least.

Update 2: probably need to bump it to an even later version **0.60.0**? since it can't parse the JSON schema refs for some reason on earlier versions, now there's some naming conflicts as its trying to export multiple `Op` and `RecordUnknown`. I think the JSON schema generation needs some updates as well.

Update 3: I believe the issue is due to the typespec, take this snippet from `output-unpacked` as an example:
```typespec
@extension("title", "Op")
enum Op {
    Insert: "i",
    Update: "u",
    Delete: "d",
    ControlSignal: "x",
    ReloadHeader: "h",
    ReloadItem: "r"
};

@discriminator("op")
@extension("title", "Notification")
model Notification {
    "op": Op;
    ts_ms: int64;
    metadata?: Record<unknown>
}

@extension("title", "ChangeNotification")
model ChangeNotification extends Notification {
    "op": Op.Insert | Op.Update | Op.Delete;
    payload: ChangePayload;

    @doc("The sequence number of the source change")
    seq: int64;
}
```
We have an union of **some** `Op` members in `ChangeNotification`, so when we generate the JSON schemas, `ChangeNotification` defines its own `Op` since it doesn't know how else to narrow the type(?):
```jsonschema
$schema: https://json-schema.org/draft/2020-12/schema
$id: ChangeNotification.yaml
type: object
properties:
  op:
    anyOf:
      - type: string
        const: i
      - type: string
        const: u
      - type: string
        const: d
  payload:
    $ref: ChangePayload.yaml
  seq:
    type: integer
    description: The sequence number of the source change
required:
  - op
  - payload
  - seq
allOf:
  - $ref: Notification.yaml
title: ChangeNotification
```

The result is something like this, which results in the conflict:
```python
# generated by datamodel-codegen:
#   filename:  ChangeNotification.yaml
#   timestamp: 2026-06-22T22:15:15+00:00

from __future__ import annotations

from enum import Enum
from typing import Annotated

from pydantic import Field

from .ChangePayload import ChangePayload
from .Notification import Notification


class Op(Enum):
    i = 'i'
    u = 'u'
    d = 'd'


class ChangeNotification(Notification):
    op: Op
    payload: ChangePayload
    seq: Annotated[int, Field(description='The sequence number of the source change')]
```
```python
# generated by datamodel-codegen:
#   filename:  Op.yaml
#   timestamp: 2026-06-22T22:15:15+00:00

from __future__ import annotations

from enum import Enum


class Op(Enum):
    i = 'i'
    u = 'u'
    d = 'd'
    x = 'x'
    h = 'h'
    r = 'r'

```
To get around this, it does some weird renaming in `__init__.py` which is really ugly:
```python
# generated by datamodel-codegen:
#   filename:  json-schema
#   timestamp: 2026-06-22T22:15:15+00:00

from __future__ import annotations

from .ChangeNotification import ChangeNotification
from .ChangeNotification import Op as ChangenotificationOp
from .ChangePayload import ChangePayload
from .ChangePayload import RecordUnknown as ChangepayloadRecordUnknown
from .ChangeSource import ChangeSource
from .ControlPayload import ControlPayload
from .ControlSignalNotification import ControlSignalNotification
from .Notification import Notification
from .Notification import RecordUnknown as NotificationRecordUnknown
from .Op import Op as OpOp
from .ReloadHeader import ReloadHeader
from .ReloadItem import ReloadItem
from .Versions import Versions

__all__ = [
    "ChangeNotification",
    "ChangePayload",
    "ChangeSource",
    "ChangenotificationOp",
    "ChangepayloadRecordUnknown",
    "ControlPayload",
    "ControlSignalNotification",
    "Notification",
    "NotificationRecordUnknown",
    "OpOp",
    "ReloadHeader",
    "ReloadItem",
    "Versions",
]
```
@amansinghoriginal I'm thinking the typespec needs some breaking changes (should be fine since its in development) if we want the Pydantic end-user models to be remain autogenerated.

Also, the .NET SDK has separate namespaces for packed models and view service models (but not unpacked models, I believe the `PostDaprPubSub` reaction is responsible for "unbatching" and publishing unpacked events). IMO having separate modules for the Pydantic models would be cleaner than having one massive `models` interface to import models from (i.e. `from drasi.reaction.models.unpacked import ...`, `drasi.reaction.models.packed import ...`), WDYT?

amansinghoriginal:
Thanks for digging into this @JeffreyJPZ, your diagnosis was pointing the right way. The real cause of the import failure is that we are generating three schemas (query-output, view-svc, output-unpacked) into one flat models directory, so the cross references resolve to modules instead of classes, and on top of that we get the Op / RecordUnknown / Versions name clashes. Chasing it through datamodel-codegen export flags and version bumps is fighting the symptom rather than the cause, so let me reframe where this is actually headed.

The thing that needs the unpacked models is the drasi trigger extension over in dapr/dapr-agents, which is a separate repo, so it cannot import from this PR anyway. For that consumer the cleanest move is to generate just output-unpacked, on its own, into its own directory inside the extension.

One schema, one `datamodel-codegen` run, one self consistent init, so the collision problem disappears because there is nothing to collide with.

JeffreyJPZ:
Closing this for now in favour of the other approach

Issue #435, opened 2026-06-18
[Tracking issue] Dapr Agents Extension


JeffreyJPZ:
This is the tracking issue for Drasi-related tasks in a Dapr Agents extension, allowing durable agents to be enriched in near real-time with business events produced by Drasi.

Project context: https://summerofcode.withgoogle.com/programs/2026/projects/qbtyohoJ
Dapr Agents: https://docs.dapr.io/developing-ai/dapr-agents/
Dapr Agents tracking issue: dapr/dapr-agents#630
- #436
- drasi-project/learning#110
- #443

PR #443, opened 2026-08-12
[WIP] Dapr agent router

JeffreyJPZ:
# Description

Add an "agent-aware" router reaction for Drasi
- Publish messages to agents listening on Dapr pub/sub topics
- Embedded MCP server supports query subscription, query unsubscription, and query listing capabilities as tools
- Durable subscription state backed by Dapr state store

Primarily concerned with Dapr Agents, but can be expanded to be framework-agnostic in the future.

Made some changes to the Python SDK so before this gets merged I need those changes and also need to drop the relevant commits here.

## Out of Scope
- Subscription refresh semantics (clients must explicitly unsubscribe)
- Security (agent identity, payload signing, multi-tenancy, etc.)

Part of #435
Related to #442

## TODO:
- [x] ~~Decide whether "Pubsub router" is an appropriate name~~ Changed to "Agent Router"
- [x] ~~Decide who should be responsible for generating agent tasks - should the reaction define tasks or publish raw event payloads and have agent developers and/or LLMs decide?~~ Event envelope should be given in the tool docstring and LLMs should generate their own instructions for handling events
- [x] ~~Decide on Drasi operation terminology for fine-grained subscriptions - "added", "updated", "deleted"?~~ Should be fine
- [x] ~~What should the "ID" arg be called: `client_id` vs `consumer_id` vs `agent_id`?~~
- [x] ~~Write-through strategy with in-memory state and Dapr state store? How would this work with multiple replicas? Do we need configuration subscriptions?~~ Only allow a single replica for now
- [x] Unit + integration tests
- [ ] Address some remaining TODOs
- [x] Read state store name and other settings(?) from env when #442 lands
- [ ] Changes to Python SDK merged + drop SDK commits from PR


reactions/dapr/agent-router/src/agent_router/mcp_server/tools.py
```diff
@@ -0,0 +1,222 @@
...
+    ) -> UnsubscribeResult:
+        """
+        Unsubscribe an agent from a Drasi query.
+
+        Args:
+            query_id (str): Unique identifier of the query to unsubscribe from.
+            agent_id (str): Unique identifier of the agent to unsubscribe.
+                Injected by the agent framework — must never be exposed to the LLM through the tool schema.
+            subscription_id (str): Unique identifier of the subscription to remove.
+                Injected by the agent framework — must never be exposed to the LLM through the tool schema.
```
JeffreyJPZ: This is quite annoying especially with fine-grained subscriptions - need to retrieve subscription ID based on query ID, event_type client-side or something similar

reactions/dapr/agent-router/src/agent_router/mcp_server/tools.py
```diff
@@ -0,0 +1,222 @@
...
+        logger.info("DrasiQueryToolSet initialized and tools registered with MCP")
+
+
+    async def subscribe(
+        self,
+        query_id: str,
+        event_types: Annotated[list[EventType], Field(min_length=1)],
+        agent_id: str,
+        topic: str,
+    ) -> SubscribeResult:
```
JeffreyJPZ: Since subscribe is an upsert, it could be useful to support an optional `subscription_id` in case of replay/scaling back up from zero.

@amansinghoriginal any thoughts?

PR #454, opened 2026-09-16
Ensure result sequence durability

### LINKED ISSUE #451: Query-host can reuse published sequence numbers after a restart
Query-host's `SequenceManager` advances its counter in memory and persists it through a background task. Publication does not wait for that checkpoint, and persistence failures are only logged.

Consequently, query-host can publish an event with sequence `101`, crash before saving the allocation, restore the previous counter of `100`, and allocate `101` to a different event. This can happen even when the query's durable storage survives the restart.

Query-host should confirm that a sequence allocation is persisted before publishing an event using it. Failed or unconfirmed persistence must prevent the dependent publication and, during normal change processing, successful acknowledgment of the source input. Persisted allocations whose events were never published may leave gaps; they should not be reclaimed.

amansinghoriginal:
Fixes #451


query-container/query-host/src/query_worker.rs
```diff
@@ -814,46 +844,105 @@ async fn deprovision_result_view(
...
+            .apply_sequence(sequence, source_change_id)
+            .await?;
 
-                if let Err(err) = store
-                    .apply_sequence(latest.sequence, &latest.source_change_id)
-                    .await
-                {
-                    log::error!("Error applying sequence: {err}");
-                }
+        self.value.sequence = sequence;
```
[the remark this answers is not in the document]
amansinghoriginal: Fixed in `bab588c`. Failed allocations are no longer reused by later events.

query-container/query-host/src/query_worker.rs
```diff
@@ -520,6 +538,7 @@ async fn process_change(
         }
     };
 
+    let seq = seq_manager.increment(&source_change_id).await?;
```
danielgerlag: This used to live inside `if !changes.is_empty()`, so a sequence was only allocated when something was actually published.

Pulling it up here burns a number on every source change, including no-ops, and the worker still acks. Downstream then sees a gap for a result that never existed.

#451 allowed gaps after persist-then-crash, when the source was not acked. This makes gaps the happy path.

Put `increment` back inside the publish branch and await it there. That is what the first commit already did.

danielgerlag (dismissed):
Sequence allocation now runs on the no-op path. That creates result-stream gaps for events that are never published.

query-container/query-host/src/query_worker.rs
amansinghoriginal: Thanks, I've moved it back in `8e5f11f`. I checked the consumers and they do tolerate gaps, but keeping allocation tied to actual output makes sense here.

I originally moved it to avoid losing a result if persistence failed after evaluation. That recovery issue is separate from sequence durability, so I'll leave it out of this PR.

